PR #1, opened 2021-06-04
Switch to yarn workspaces

ice9js:
Yarn workspaces provide native monorepo support and tus a slightly better experience when dealing with shared dependencies. Hence the switch.

I also added a doc describing the to-be directory structure and how to work with it.


CGastrell (approved):
LGTM!

PR #2, opened 2021-06-04
Add @crowdsignal/http

ice9js:
Moves our frontend http wrapper into its own package. I also went ahead and added a proper `README.md` file.


CGastrell (approved):
Who am I to oppose? :shipit:

PR #3, opened 2021-06-04
Add @crowdsignal/rest-api

ice9js:
This patch adds a new `@crowdsignal/rest-api` package which will serve as an abstraction layer for our `api.crowdsignal.com` API.

The idea is this package would be used to transform any data going between the app and the API in order to improve or expand compatibility.

Depends on #2.


CGastrell (approved):
Ok

PR #4, opened 2021-06-04
Add @crowdsignal/styles

ice9js:
Adds a new `@crowdsignal/styles` package meant to provide all our styling utilities in JS form to support styled components.  
As a starting point, I migrated our color palette, added a lightweight function for generating breakpoints and a very minimal reset sheet.  
More will be migrated as required.


CGastrell (approved):
Might come in handy :shipit:

PR #5, opened 2021-06-04
Add @crowdsignal/components

ice9js:
This package will be the host of our main shared component library.  
For a start, I added `<CrowdsignalFooter>` and `<StyleProvider>` which are required by the feedback widget.

In addition, I also migrated our old `<Button>` and `<FormTextInput>` components, but decided not to use them in the widget after all, as it has its own distinct look and making it look like the dashboard could take away from it.  
But as they're already moved over, it doesn't hurt to have them included.

Depends on #4.

### DIFF
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -1,0 +1,3 @@
+# Crowdsignal components
+
+The main frontend components directory for our apps.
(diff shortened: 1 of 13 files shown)


CGastrell (approved):
No blocks? :shipit:

PR #6, opened 2021-06-04
Add @crowdsignal/feedback-widget

ice9js:
This adds a stand-alone implementation of our Feedback Button widget from Crowdsignal Forms.

**Note:** The feedback button block relies on `<Dropdown>` from `@wordpress/components` for the popover. I've had to manually copy the CSS for that for the component to work correctly here although it's still not 100% ideal. As a long term solution we should implement our own version of that and avoid relying on `@wordpress/components` outside of block editor.

**Note:** The settings are temporarily hardcoded as a `settings` object inside `widget.js` until we have proper API support for them.

Depends on #3 and #5.
 
# Testing

- Run `yarn workspace @crowdsignal/feedback-widget run build` to build the widget.
- Run `yarn workspace @crowdsignal/feedback-widget run build:wp` to build a cut-down version of the widget for WordPress sites. Without including libraries that WordPress already provides.
- Test the widget by including the script on a page and creating a `FeedbackWidget` through either a script or manually.

### DIFF
--- a/apps/feedback-widget/README.md
+++ b/apps/feedback-widget/README.md
@@ -1,0 +1,14 @@
+# Feedback widget
+
+Embed the feedback widget on your page and collect feedback with Crowdsignal.com.
+
+## Usage
+
+Add the following JS snippet to enable the feedback widget to your page. Make sure to replace `surveyId` with a valid feedback survey ID.
+
+```html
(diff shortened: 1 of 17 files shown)


CGastrell (approved):
Looks good, let's work with this :D

PR #11, opened 2021-07-07
Add @crowdsignal/router

ice9js:
This patch adds and implements a new `@crowdsignal/router` which will be used for efficient navigation throughout the new Crowdsignal frontend, while still allowing to maintain compatibility with legacy views.

For the exact description of the API, see `README.md`.

# Testing

Testing instructions can be found in  d63837-code.

### DIFF
--- a/packages/router/src/router.js
+++ b/packages/router/src/router.js
@@ -1,0 +1,50 @@
+/**
+ * External dependencies
+ */
+import { useCallback, useEffect, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { RouterContext, history } from './context';
+import { getLinkElement, hasLocalTarget, parseLocation } from './util';
+
+export const Router = ( { children, allowedRoutes = /.*/i } ) => {
+	const [ route, setRoute ] = useState( parseLocation( history.location ) );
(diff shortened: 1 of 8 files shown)


packages/router/src/router.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+import { useCallback, useEffect, useState } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { RouterContext, history } from './context';
+import { getLinkElement, hasLocalTarget, parseLocation } from './util';
+
+export const Router = ( { children, allowedRoutes = /.*/i } ) => {
```
CGastrell, line 12: is `allowedRoutes` a single regex for all? Could/should we turn this into an array?

CGastrell (approved):
Seems like we'll be doing things at home, sweet router. I just wondered about the `allowedRoutes`, but seems like a really good start :)

packages/router/src/router.js
ice9js: I guess we could... that's a good point.  
Let's maybe open up an issue for that or something just to get the initial things merged so it can easily be tested when we alter it.

PR #12, opened 2021-07-07
Add @crowdsignal/hooks

ice9js:
This patch creates a new package called `@crowdsignal/hooks` which should serve as a collection for any of the custom hooks shared between our components.

Initially, I moved `useOnOutsideClick` - previously defined in `Popover` - into the new package and renamed it to `useBlur` so it can easily be used by other components as well.

# Testing

The new package and hook can be tested by running `yarn storybook` and verifying the `Popover` component is still working as it did.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -2,23 +2,22 @@
   "name": "@crowdsignal/components",
   "version": "0.1.0",
   "description": "Crowdsignal frontend component library.",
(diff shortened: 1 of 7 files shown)


CGastrell:
Code looks good, but there was no `Popover` explicit demo on `yarn storybook`. The popovers/dropdown menus on the editor worked, is that enough as proof?

PR #14, opened 2021-07-07
Update @crowdsignal/styles reset stylesheet

ice9js:
This patch is a quick fix for our reset stylesheet which adjusts the next two things:

- Use `:host` when injecting the stylesheet into a shadow DOM node, or `:root` otherwise.
- Inject our entire color palette into the stylesheet as CSS variables. This makes for more performant css-in-js styling and enables regular CSS - or SCSS - to be used as well.

This is not the endgame quite yet. I'm thinking about splitting the responsibilities between injecting variables and the reset sheet so it's cleaner and we have more control when applying these in our widgets.  
In the mean time, it does the job.

# Testing

You can try loading storybook and verifying the CSS color variables show up as expected and the default styles are applied.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -6,7 +6,7 @@ import { Global } from '@emotion/core';
 /**
  * Internal dependencies
  */
-import { StyleProvider } from '@crowdsignal/components';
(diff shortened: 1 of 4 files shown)


CGastrell (approved):
LGTM!

PR #15, opened 2021-07-07
Implement a TabNavigation component

ice9js:
This patch implements a `<TabNavigation>` component which will be used as secondary navigation throughout Crowdsignal.

Screen Shot 2021-07-07 at 11 05 38 PM

# Testing

The new component can be seen and tested in storybook.


CGastrell:
Not seeing the same output:
image

And clicking doesn't switch, it brings up the fullscreen story, is this expected?

CGastrell (approved):
IPF :shipit:

PR #17, opened 2021-07-07
Implement PageHeader and EditablePageHeader components

ice9js:
This patch implements a `<PageHeader>` and an `<EditablePageHeader>` components for displaying page headers across the dashboard in both editable and non-editable forms.

Screen Shot 2021-07-07 at 11 26 49 PM

# Testing

Both components can be seen and tested inside storybook.


CGastrell (approved):
IPF :shipit:

PR #20, opened 2021-08-04
Add development server configuration for @crowdsignal/dashboard

ice9js:
This patch adds a simple webpack dev server configuration for Crowdsignal dashboard to facilitate local development.

The local server can be ran with `yarn workspace @crowdsignal/dashboard start` and will be available under `crowdsignal.localhost:9000`.

The last part is important - it requires an entry in `/etc/hosts`, but has the added advantage over using plain `localhost` that the browser will allow CORS when talking to our API.

# Testing

- Apply the patch and run `yarn install`.
- Add `127.0.0.1 crowdsignal.localhost` to `/etc/hosts`
- Run `crowdsignal.localhost:9000`
- Try accessing `https://crowdsignal.localhost:9000/edit/poll/123` to see the block editor.

### DIFF
--- a/apps/dashboard/babel.config.js
+++ b/apps/dashboard/babel.config.js
@@ -1,0 +1,10 @@
+module.exports = {
+	presets: [
+		'@automattic/calypso-build/babel/wordpress-element',
+		'@automattic/calypso-build/babel/default',
+	],
+	plugins: [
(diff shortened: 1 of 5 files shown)


CGastrell (approved):
LGTM!

PR #21, opened 2021-08-06
Implement a `redirect()` method for @crowdsignal/router

ice9js:
This patch implements a simple `redirect()` method which will serve as a means to programmatically switch the user to a different route/view inside the app.

For example

```javascript
import { redirect } from '@crowdsignal/router';

redirect( '/' ); // Will redirect to the root.
```

# Testing

No tests 🤷


CGastrell (approved):
IPF!

PR #22, opened 2021-08-06
Add missing color values

ice9js:
While working on other changes, I noticed some of the color values from the palette are missing in our code hence I opted to re-add them where necessary.

# Testing

If you build the app, the added colors should be injected onto the page as CSS variables.


CGastrell (approved):
Moar colors! 🌈

PR #23, opened 2021-08-06
Add poll & project routes to @crowdsignal/rest-api

ice9js:
This patch adds new poll and project route definitions to the `@crowdsignal/rest-api` package.  
As the payloads haven't been finalized yet, the README files will need to be updated later.

Update requests depend on d65222-code.

# Testing

Currently, it's difficult to test these changes on their own.

### DIFF
--- a/packages/rest-api/src/index.js
+++ b/packages/rest-api/src/index.js
@@ -1,1 +1,3 @@
 export * from './feedback';
+export * from './poll';
+export * from './project';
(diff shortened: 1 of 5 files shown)


CGastrell (approved):
In the end, if we go with the `projects` idea, maybe the `polls` fetchers would not be needed, but good to have them while we properly shape this.

PR #24, opened 2021-08-06
Add dashboard state

ice9js:
This patch adds the initial state for the dashboard. So far I've added the following:

- Actions & controls for saving polls and projects to the Crowdsignal backend.
- A `redirect` action with a control to trigger `redirect()` from `@crowdsignal/router`.
- A complete implementation of state for notices, that will be used to display succes/info/error notices throughout the app - however we don't have a notice component just yet.

Depends on #21 and #23.

# Testing

These changes are pretty much internal and remain unconnected at this point. No tests necessary.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,5 +31,12 @@
 		"@wordpress/no-unused-vars-before-return": [
 			"off"
 		],
+	},
+	"settings": {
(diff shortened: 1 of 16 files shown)


CGastrell (approved):
IPF!

PR #25, opened 2021-08-06
Refactor <PollEditor /> into a unified project editor

ice9js:
This patch renames PollEditor and all associated components into a generic `<Editor>`.  
I've also adjusted the saving logic according to our latest decision, to deal with projects rather than polls.

For now, the saving logic is a simple `debounce` where it will attempt to save your changes a second after you stop typing.
I believe this should serve its purpose well initially, and we can implement a proper buffering solution later down the line.  
By doing it that way, we should also have a better idea of whether we want to buffer each part of the project separately or if we want to aggregate and debounce all changes to any project fields together.

Depends on #24.

# Testing

- When you try to go to `/edit/poll`, the editor will attempt to create a new project which will fail due to API currently returning `501 Not Implemented`. You should be redirected to `https://app.crowdsignal.com`.

- When you go to `/edit/poll/123` editor should load normally, fetching logic isn't wired up yet.

- When you type for a bit and stop, you should observe a request go out to `/v4/projects/:projectId` to save the changes.
As of right now, the request fails with `501 Not Implemented.


CGastrell (approved):
LGTM!

PR #26, opened 2021-08-11
Create @crowdsignal/blocks package

ice9js:
This patch adds a new @crowdsignal/blocks package and can serve to open the debate on how to organise things a little in terms of sharing block code between the dashboard and the plugin.

Unfortunately, with the child blocks, the blocks themselves do have a hardcoded dependency on their child block 'id' so first of all we need to use the same prefix (`crowdsignal-forms/` ) for all our block in the plugin and the dashboard.

The way it's done now, the package exposes complete blocks ready to be registered. But we could even go one step further and include the register function in the package itself as well and use `allow/disallowBlocks` to restrict the editor in each respective app. Perhaps this is even a better approach?

Finally, my last idea: because the blocks will have a somewhat different saving flow on the platform and in WordPress: on the platform we'd only create the poll once you publish but in WordPress as soon as you add/change the block.
I'm thinking we should probably solve this by using a higher order component that injects the appropriate handler (or `noop` if not necessary), and then use compile-time text substitution to include the correct one.

# Testing

- Run `yarn workspace @crowdsignal/dashboard start` and go to `https://crowdsignal.localhost:9000/edit/poll/123`.
- Try adding a `Poll` block
- A poll block should appear preconfigured with three poll answers and you should be able to add more answers and paragraphs inside it.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -19,10 +19,12 @@
   },
   "main": "src/index.js",
   "dependencies": {
+    "@crowdsignal/blocks": "^0.1.0",
     "@crowdsignal/block-editor": "^0.1.0",
     "@crowdsignal/components": "^0.1.0",
     "@crowdsignal/rest-api": "^0.1.0",
     "@crowdsignal/router": "^0.1.0",
+    "@wordpress/blocks": "^9.1.4",
     "@wordpress/data": "^6.0.0",
     "@wordpress/element": "^4.0.0",
     "@wordpress/i18n": "^4.2.1",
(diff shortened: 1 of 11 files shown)


CGastrell:
Had some issues trying to test this:
image

Wanna do an online session?

CGastrell:
> Unfortunately, with the child blocks, the blocks themselves do have a hardcoded dependency on their child block 'id' so first of all we need to use the same prefix (crowdsignal-forms/ ) for all our block in the plugin and the dashboard.

I'm ok with it. In fact, it kind of make sense not use simply `crowdsignal` all over the place.

ice9js:
This should work now - don't forget to `yarn install` before testing.

CGastrell (approved):
Beautiful

PR #27, opened 2021-08-13
Send request with blocks

CGastrell:
This PR changes the save handler on the editor so the raw blocks can be sent instead of the HTML markup.

`onSaveBlocks` accepts a second argument which represents *ignoredContent*. This argument carries all things NOT to be parsed, like new lines and the kind. For now, it's just ignored (well thought argument name here, kudos).

Also added the payload on the `rest-api.updateProject`  call.

Ref: [onSaveBlocks](https://github.com/Automattic/isolated-block-editor#onsaveblocks)

### DIFF
--- a/apps/dashboard/src/data/projects/controls.js
+++ b/apps/dashboard/src/data/projects/controls.js
@@ -10,6 +10,6 @@ export default {
 			return createProject( project );
 		}
 
-		return updateProject( projectId );
+		return updateProject( projectId, project );
 	},
 };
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/data/projects/controls.js
```diff
@@ -10,6 +10,6 @@ export default {
 			return createProject( project );
 		}
 
-		return updateProject( projectId );
+		return updateProject( projectId, project );
```
ice9js, line 13: Thanks!

ice9js (approved):
Looks good to me! 👍

PR #28, opened 2021-08-16
Work editor flow

CGastrell:
This PR tests a non-automagic saving editor, adding 2 simple yet needed action buttons: Save draft and Publish.

### MAYBE NOT FOR MERGE

The changes explore how to make for a full editor flow: new project, save (switch fetching state), maybe update trigger a router path update, and finally updateProject (which doesn't actually triggers any request, only updates the state)

@ice9js : not entirely convinced by the namings and still working out why the initial state doesn't show on devtools or why the router is not updating the URL. Also, the project action seems to be taking care of every single step of the flow, but working so far (against D65459-code)

Commented code is to be reinstated once we figure this out manually


apps/dashboard/src/components/project-navigation/style.scss
```diff
@@ -4,7 +4,6 @@
 	display: grid;
 	grid-template-columns: 1fr auto 1fr;
 	padding: 0 32px;
-	width: 100%;
```
ice9js, line 7: I believe this is actually quite important for the buttons to stay centered and not to move when you edit the title (and not be affected by the temporary buttons either).

apps/dashboard/src/components/project-tools/index.js
```diff
@@ -0,0 +1,49 @@
...
+				{ __( 'Save draft', 'dashboard' ) }
+			</Button>
+			<Button primary onClick={ publishProject } disabled={ isFetching }>
+				{ __( 'Publish', 'dashboard' ) }
+			</Button>
+		</div>
+	);
+};
+
+export default withSelect( ( select ) => {
```
ice9js, line 45: Nit: I think `useSelect` is the preferred/cleaner way of doing it nowadays. It's what `withSelect` does under the hood anyway.

That said,... this whole component is kind of temporary until we implement autosave so feel free to disregard the comment and keep in mind for future stuff ;)

apps/dashboard/src/data/action-types.js
```diff
@@ -5,3 +5,5 @@ export const POLL_UPDATE = 'POLL_UPDATE';
 export const PROJECT_SAVE = 'PROJECT_SAVE';
 export const PROJECT_UPDATE = 'PROJECT_UPDATE';
 export const ROUTE_UPDATE = 'ROUTE_UPDATE';
+export const FETCH_ONGOING = 'FETCH_ONGOING';
```
ice9js, line 8: This is pretty bad naming. Like, even in the beta state we'll already have scenarios with multiple requests happening at the same time. And not all of them are made equal hence these action names should be specific to the resource they affect.

Also, I never added the comment here since we've been doing it on the dashboard but it helps to follow a similar naming scheme for all of these, that being: `RESOURCE_NAME_ACTION`. Hence for example `ROUTE_UPDATE` being a thing instead of just `REDIRECT`.
And alphabetical order please.

`PROJECT_SAVE` was intended for this very purpose in this scenario - and as such made separate from `PROJECT_UPDATE`. Likewise you'd use something like `PROJECT_FETCH` if you're just retrieving an existing project for editing or otherwise.

Makes it much easier to browse and you can bet on this file growing a lot in the future.

apps/dashboard/src/data/index.js
```diff
@@ -12,6 +12,9 @@ import pollControls from './polls/controls';
 import projectControls from './projects/controls';
 import * as pollSelectors from './polls/selectors';
 import * as projectSelectors from './projects/selectors';
+import * as uiActions from './ui/actions';
```
ice9js, line 15: Lol, did I really just... not add these? Thanks!

apps/dashboard/src/data/ui/actions.js
```diff
@@ -53,3 +59,15 @@ export function redirect( path ) {
 		path,
 	};
 }
+
+export function fetchOngoing() {
```
ice9js, line 63: Like I said, very bad idea to have a single state branch to handle all possible requests. This belongs in the project state, where we should probably add `saving` and `loading` keys.

apps/dashboard/src/data/projects/actions.js
```diff
@@ -21,13 +22,21 @@ export function updateProject( projectId, project ) {
...
 
-		return updateProject( id, {
+		yield updateProject( id, {
 			...project,
 			id,
 		} );
+
+		if ( ! projectId ) {
+			// this should only happen when creating a new project
+			yield redirect( `/${ id }` );
```
ice9js, line 36: Like I mentioned in our last conversation, the redirect doesn't really belong in the state. I mean, it would work, but it'd be better to do it from the component since here you don't have the context where you want to redirect to or why.

While it works, if we ever needed to dispatch this action from anywhere else than the editor for whatever reason this will become a problem. Hence in the original version it was part of the `Editor` component.

apps/dashboard/src/data/projects/actions.js
```diff
@@ -21,13 +22,21 @@ export function updateProject( projectId, project ) {
...
+		}
+
+		return fetchIdle();
 	} catch ( error ) {
 		// Save failed
```
ice9js, line 41: 1st of all, if the request fails we're now hanging in a permanent `saving = true` state. You'll need to add a `PROJECT_SAVE_FAILED` / `PROJECT_SAVE_ERROR` / or alike action to handle that case too.

Might as well rename the fetch idle to `PROJECT_SAVE_SUCCESS` then.

ice9js (changes requested):
Left some comments regarding the 'fetching' state for now.

apps/dashboard/src/components/project-navigation/style.scss
CGastrell: Fair enough, but we might need to find a better solution, maybe a flex-grow or something as it makes the Navigation container to overflow the viewport

apps/dashboard/src/components/project-tools/index.js
CGastrell: TBH, I initially tried with useSelect but it didn't do the job. That said, I was doing so many changes that maybe my initial use of `useSelect` was flawed. Will try again.

apps/dashboard/src/data/action-types.js
CGastrell: Yeah, I hear you. I was thinking on a single on/off thing in case (eventually) we found ourselves doing independent and separate requests (title comes to mind), so having a single action that just switches the *fetching state* (hence, I created the action creator on `ui`) made sense. But I agree with the usual use of this and will change it.

ice9js: Yeah... if we have separate request, then it might actually be best to manage that part of the state separately as well. Just to avoid dealing in any kind of race conditions at all.

apps/dashboard/src/data/projects/actions.js
CGastrell: Not entirely convinced about the route not being part of the state. It is something that actually *changes* and it makes for different behaviors. If you think of it, it's very similar to `lastUpdatedItemId`: defines what you're working against and is something on the *editor* (`ui`) state. I'm ok sticking with `lastUpdatedItemId`, but a router state that states what route (project) I'm working on makes more sense to me.
Thinking on switching from editor to results: the route changes and we'll still need some selector to know where are we and what we're working against.

CGastrell:
Still saving a new one each time (update not taking projectId)

apps/dashboard/src/components/editor/index.js
```diff
@@ -76,4 +57,10 @@ const Editor = ( { projectId } ) => {
 
 registerBlocks();
 
-export default Editor;
+export default withSelect( ( select, ownProps ) => {
```
ice9js, line 60: Side note: hooks are generally preferred to HoC's as they're slightly more readable and performant.  
However, a good use of HoC's would be when you want to export the 'bare' component as well as a version connected to the store - like with the `<Masterbar />` for whatever purpose.

Anyway, let's keep this now but use hooks when possible in the future for consistency.

apps/dashboard/src/data/ui/controls.js
```diff
@@ -1,11 +1 @@
...
- */
-import { redirect } from '@crowdsignal/router';
-import { ROUTE_UPDATE } from 'data/action-types';
-
-export default {
-	[ ROUTE_UPDATE ]: ( { path } ) => {
-		redirect( path );
-	},
-};
+export default {};
```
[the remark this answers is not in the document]
ice9js: I still stand by the decision to remove - as in delete - this file entirely, never to return.

apps/dashboard/src/data/ui/actions.js
```diff
@@ -53,6 +54,7 @@ export const infoNotice = ( message, { id, ...options } ) =>
 	createNotice( id || uniqueId(), 'info', message, options );
 
 export function redirect( path ) {
```
ice9js, line 56: We won't be needing this action anymore either, let's just remove it along with its action type.  
Just as you did here, you can call `routeRedirect()` directly _anywhere_ hence no need to do the whole wp.data dance for it.

apps/dashboard/src/data/projects/controls.js
```diff
@@ -1,11 +1,46 @@
 /**
  * Internal dependencies
  */
-import { createProject, updateProject } from '@crowdsignal/rest-api';
-import { PROJECT_SAVE } from '../action-types';
+// import { createProject, updateProject } from '@crowdsignal/rest-api';
+import { PROJECT_SAVE_REQUEST } from '../action-types';
```
ice9js, line 5: The comment got hidden after the rebase so re-adding it here so you don't miss it:

#33 added the `dispatchAsync` action/control which should make this obsolete.

As in, instead of dispatching `PROJECT_SAVE_REQUEST`, in your action generator, you can just:

```javascript
// Notice these are the request functions themselves, not actions
import { createProject, updateProject } from '@crowdsignal/rest-api';

...

export function* saveAndUpdateProject( projectId, project ) {

    // Dispatch PROJECT_SAVE, now used only for toggling the `saving` state.
    yield saveProject( projectId );

    try {
        // Use dispatchAsync to trigger the actual request now
        const response = yield projectId
            ? dispatchAsync( updateProject, [ projectId, project ] )
            : dispatchAsync( createProject, [ project ] ) ;

        // Any specific post processing of the response can happen here if necessary

        // Finally, update the project.
        // If we work under the assumption it's impossible for the project to update while it's saving
        // which sounds reasonable, then PROJECT_UPDATE can reset the saving flag as well which makes PROJECT_SAVE_SUCCESS obsolete.
        yield projectUpdate( response.data.id, response.data );
    } catch ( error ) {
        // Request failed
        yield saveProjectError( projectId );
    }
}
```

In fact, `dispatchAsync` should literally remove the need for any other control as it already allows for calling any asynchronous function from an action generator or resolver - the main purpose of controls - with minimal pollution to the store namespace.

ice9js (changes requested):
Left a few more comments.

apps/dashboard/src/data/projects/actions.js
```diff
@@ -19,17 +24,32 @@ export function updateProject( projectId, project ) {
...
-			id,
-		} );
+		yield updateProject( id, { ...project, id } );
+		yield redirect( `/edit/poll/${ id }` );
+		return saveProjectSuccess( id );
 	} catch ( error ) {
-		// Save failed
-		throw error;
+		// eslint-disable-next-line
+		console.error( error );
```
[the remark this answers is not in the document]
CGastrell: `debug` lib?

PR #29, opened 2021-08-17
Bump isolated-block-editor to version 2.3.0

ice9js:
This patch bumps the version of `isolated-block-editor` to `2.3.0` which adds a `<ToolbarSlot />` component that allows for inserting items into the editor toolbar.

As part of this PR, I also manually bumped all our dependencies on `@wordpress/...` libraries to match the ones used in `isolated-block-editor`.  
This solves the issue with multiple versions of the same library being installed causing issues for our custom blocks (see #26). Eventually we'll want to make them external altogether but for now this will do.

Another change is how now the styles need to be imported manually.

# Testing

- Run `yarn install`
- Run `yarn workspace @crowdsignal/dashboard start` and go to `https://crowdsignal.localhost:9000/edit/poll/123`
- The editor should still load normally with the correct styles.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -24,8 +24,8 @@
     "@crowdsignal/rest-api": "^0.1.0",
     "@crowdsignal/router": "^0.1.0",
     "@wordpress/data": "^6.0.0",
-    "@wordpress/element": "^3.1.1",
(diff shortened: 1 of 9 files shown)


CGastrell (approved):
Perfecto! We'll deal with the 2 offending `width: 100%` later :)

PR #30, opened 2021-08-18
List peer dependencies explicitly

ice9js:
This patch adds explicit mentions of all required peer dependencies into `devDependencies` in the root `package.json` which does away with the wall of warnings every time we run `yarn install`, which should help in case something actually does go wrong.

I based the version numbers on the ones that were already installed (albeit implicitly) using `yarn list`.

I managed to get rid of all but two warnings:

- The latest calypso-build has a dependency on `cache-loader@4.1.0` which has a peer dependency on `webpack@^4.0.0` but `calypso-build` itself has a peer dependency on `webpack@^5.46.0`. (We use webpack 5).
- `@wordpress/url` depends on a polyfill which has a peer dependency on `react-native`. We don't actually use `react-native` at all and installing it causes some more, harder to resolve issues so leaving it for now at least.

# Testing

- Run `yarn install`, everything should install correctly and there should only be two warnings now.
- Run `yarn workspace @crowdsignal/dashboard start` to see that everything still works correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,14 +14,23 @@
   },
   "devDependencies": {
     "@automattic/calypso-build": "^7.0.0",
+    "@babel/core": "7.14.6",
     "@storybook/builder-webpack5": "^6.3.0",
     "@storybook/manager-webpack5": "^6.3.0",
     "@storybook/source-loader": "^6.3.0",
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
:shipit:

PR #31, opened 2021-08-19
Implement settings for poll and poll answer blocks

ice9js:
This patch adds settings for `Poll` and `PollAnswer` blocks based on the ones in Crowdsignal Forms albeit with some differences.

The main difference lies in `PollAnswer` now being a child block, and thus being able to contain its own settings.
This simplifies the main block itself quite a bit, with a potential downside that each answer is now managed separately.  
Now, one could just as well say it's an advantage as it gives the user more option. Pressing enter on one answer will duplicate all its attributes but `label` and `clientId` which makes it easy enough to add identically styled answers if you ask me.

Other than that, the changes are mostly internal:

- Each block generates a unique `clientId` when it's created, with the exception of `PollAnswer` when it's _split_ and already contains a value/label - it will retain its current `clientId`.
- `restrictions` are now a single `object` prop vs having all of them pollute the attributes.
- All style related props follow the naming scheme supported by core blocks and their helpers, as well as `@crowdsignal/styles`.
- I moved over the typography settings from Crowdsignal Forms as well but I believe there's something in core for that as well. I'll need to check.

Screen Shot 2021-08-19 at 10 18 40 PM

# Testing

- Run `yarn install` and `yarn workspace @crowdsignal/dashboard start`
- Go to `https://crowdsignal.localhost:9000/edit/poll/123`
- Add a `Poll` block
- The poll block should have a border by default, and poll answers should look like regular button blocks
- Try changing some of the color or border values on either the poll block or poll answers to see if attributes are updated correctly
- Resizing the poll block should work as well

### DIFF
--- a/packages/block-editor/src/index.js
+++ b/packages/block-editor/src/index.js
@@ -14,7 +14,14 @@ import Toolbar from './toolbar';
  */
 import './style.scss';
 
-const settings = {};
+const settings = {
+	iso: {
+		toolbar: {
+			inspector: true,
+			toc: true,
+		},
+	},
+};
 
 export const BlockEditor = ( { onSave } ) => {
 	return (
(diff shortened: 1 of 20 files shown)


CGastrell (approved):
Nice! This gonna be a rebase nightmare, but I deserve it

PR #34, opened 2021-08-23
Fix z-index property on the editor component

ice9js:
This patch fixes the `z-index` property on the editor component which made it un-selectable and thus impossible to add or edit blocks.

I removed the property from the editor and adjusted the `z-index` value on `<Popover />` so it doesn't get overlapped by the editor.

# Testing

- Go to `/edit/poll/123`.
- Check if the editor works normally and you can add blocks.
- The user menu in the top right corner shouldn't get overlapped by the edtor.


CGastrell (approved):
LGTM! :shipit:

PR #35, opened 2021-08-23
Fix storybook config

ice9js:
It seems our storybook config got broken somewhere in between all our other changes.  
This PR restores a working configuration:

- Storybook now uses the same dependency resolver configuration from calypso-build as our other apps.
- React has been bumped to version `17`. Matching the version included in `@wordpress/element`. It cases some extra peerDependency warnings but it's just nested dependencies that haven't been updated yet I guess.
- Reverted the default resolution of `./src` for `@crowdsignal/dashboard` which would cause issues one way or another as storybook pulls from all our apps and packages.

# Testing

- Run `yarn install`.
- Run `yarn storybook` and wait for storybook to load.
- Run `yarn workspace @crowdsignal/dashboard start` to see if our dashboard still works as expected. (React upgrade & internal dependency changes).


CGastrell (approved):
Thanks for bringing this back Kuba, LGTM!

PR #36, opened 2021-08-25
Add a poll widget

ice9js:
This is some initial work towards a poll widget applet. Right now, it takes the output from the editor and reconstructs it into a react tree - substituting our poll and poll answer components where necessary.  

While the widget doesn't do much yet by itself, this PR does include some interesting developments:

- I tweaked the editor to use `onChangeContent` instead of `onChangeBlocks` and manually `parse()` it back into blocks. This created an additional `originalContent` prop on each block which is very useful for reconstructing static blocks.
- Because of the above, I added a `debounce` instance right inside `onChangeContent` blocks itself, so `onSave` only gets triggered about one second of no activity in the editor. I think that's a sane decision as running `parse()` for the entire document on each keystroke could be quite expensive.
- I implemented a `renderBlocks` function which takes the object output from the editor and converts it into react elements. By default it relies on the `originalContent` prop but it's possible to provide custom renderers for each block type using the second argument. This enables dynamic blocks - for now just ours.
- Some blocks, like media-text have multiple children elements which makes it difficult to guess where `InnerBlocks` content should go. For each such case, we'll need to manually add the special behavior in `appendInnerBlocks`. By default, inner blocks are injected directly into the root node of the block.

Screen Shot 2021-08-25 at 1 32 58 PM

# Testing

Run `yarn storybook` and checkout the `Poll Widget` component.
Validate it's rendered correctly according to the content of `stories/example.js`.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,6 +1,6 @@
 const stories = [
-	'../apps/*/!(node_modules)/**/stories/*.js',
-	'../packages/*/!(node_modules)/**/stories/*.js',
+	'../apps/*/!(node_modules)/**/stories/!(example).js',
+	'../packages/*/!(node_modules)/**/stories/!(example).js',
 ];
 
 module.exports = {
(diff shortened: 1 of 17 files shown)


CGastrell (approved):
LGTM!

PR #41, opened 2021-09-14
Add Free Text block with inline debugging

CGastrell:
This PR adds a new question block: Free Text

It includes a `<pre>` with all the block's structure for easy debugging while we finish this up.

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`. Go to crowdsignal.localhost:9000 and add the new block by typing `/Free Text`

PR #42, opened 2021-09-14
Change editor route

CGastrell:
This PR removes the `poll` part of the editor route, becoming:

  - /edit/{id}
  - /edit/{id}/results

This PR depends on #41

### DIFF
--- a/apps/dashboard/src/data/projects/actions.js
+++ b/apps/dashboard/src/data/projects/actions.js
@@ -44,8 +44,7 @@ export function* saveAndUpdateProject( projectId, project ) {
 			: yield dispatchAsync( createProject, [ project ] );
 
 		yield updateProject( response.data.id, response.data );
-		redirect( `/project/${ response.data.id }` );
-		return;
+		return redirect( `/project/${ response.data.id }` );
 	} catch ( error ) {
 		// Request failed
 		return saveProjectError( error.message );
--- a/apps/dashboard/src/data/projects/actions.js
+++ b/apps/dashboard/src/data/projects/actions.js
@@ -43,8 +43,9 @@ export function* saveAndUpdateProject( projectId, project ) {
 			? yield dispatchAsync( patchProject, [ projectId, project ] )
 			: yield dispatchAsync( createProject, [ project ] );
 
+		yield updateProject( response.data.id, response.data );
 		redirect( `/project/${ response.data.id }` );
-		return updateProject( response.data.id, response.data );
+		return;
 	} catch ( error ) {
 		// Request failed
 		return saveProjectError( error.message );
--- a/apps/dashboard/src/data/projects/actions.js
+++ b/apps/dashboard/src/data/projects/actions.js
@@ -43,8 +43,7 @@ export function* saveAndUpdateProject( projectId, project ) {
 			? yield dispatchAsync( patchProject, [ projectId, project ] )
 			: yield dispatchAsync( createProject, [ project ] );
 
-		yield redirect( `/project/${ response.data.id }` );
-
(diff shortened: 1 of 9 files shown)


ice9js (changes requested):
Thanks for taking care of this, I'd say we can do slightly better URLs if we want to make them nice, especially `edit/{id}/results` doesn't sit quite right with me (but maybe that's just me?).

How about sticking with _somewhat restful_ principles, so `/project/:projectId/(edit|results|preview)`?
We can also replace `/project` with `/form/...` or just `/f/...` trello style if you think it looks better or want to keep our URLs short.

What do you think?

CGastrell:
> Thanks for taking care of this, I'd say we can do slightly better URLs if we want to make them nice, especially `edit/{id}/results` doesn't sit quite right with me (but maybe that's just me?).
> 
> How about sticking with _somewhat restful_ principles, so `/project/:projectId/(edit|results|preview)`?
> We can also replace `/project` with `/form/...` or just `/f/...` trello style if you think it looks better or want to keep our URLs short.
> 
> What do you think?

I think it fits! It's the way we used to do it with the other entities and it makes a lot of sense. Now for the big question: plurals

  - [ ] `/project`
  - [ ] `/projects`

Then, the main action, because I'm guessing the `/edit` prefix might get in our way.

  - [ ] `/project(s){/:projectId`
  - [ ] `/project(s){/:projectId/edit`

apps/dashboard/src/data/projects/actions.js
```diff
@@ -43,7 +43,7 @@ export function* saveAndUpdateProject( projectId, project ) {
 			? yield dispatchAsync( patchProject, [ projectId, project ] )
 			: yield dispatchAsync( createProject, [ project ] );
 
-		yield redirect( `/edit/${ response.data.id }` );
+		yield redirect( `/project/${ response.data.id }` );
```
ice9js, line 46: I believe calling that after `updateProject` not only makes more sense but also fixes the issue where the editor temporarily flickers the loading placeholder.

Also, not sure if `yield` is actually necessary anymore since this doesn't return an action anymore.

ice9js (changes requested):
`/project` it is and I'm fine with omitting `/edit` by assuming it as the default 👍 .

Now I'm just wondering about the redirect handler.

apps/dashboard/src/data/projects/actions.js
CGastrell: Maybe I'm not getting it right but, the redirect is called AFTER the `updateProject`, did you mean BEFORE?

On the other hand, the `updateProject` is actually to blame for the full re-render, but that's mostly on behalf of a flawed flow/sequence (the one I'm trying to address and, I think, out of this scope)

ice9js: Right now, it's:

1. Making the request
2. Triggering the redirect
3. Updating the state with the new data

2. and 3. need to be swapped around.

CGastrell: Done. I don't think the flickering will be gone, the problem persists and I think is elsewhere. We'll deal with it

ice9js: Not quite... what I meant was It should be like this:

```javascript
yield updateProject( ... );

redirect( ... );
return;
```

Or even just `return redirect( ... );` - it doesn't really change anything in the handling and is a little bit nicer to read. (Returning undefined means there's nothing to do.)

CGastrell: Sorry! Messed up that stash 😬

ice9js (approved):
Works great now! :ship: it

PR #44, opened 2021-09-16
Add multiple choice question block

CGastrell:
This PR adds the initial skeleton for the MC Question Block.

It is based on the work for the Poll block and should suffice while the backend is set to receive and save this type of questions.

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`, visit http://crowdsignal.localhost:9000/project and insert a MC Question Block by typing `/Multiple Choice`

### DIFF
--- a/packages/blocks/src/multiple-choice/edit.js
+++ b/packages/blocks/src/multiple-choice/edit.js
@@ -1,0 +1,61 @@
+/**
+ * External dependencies
+ */
+import { InnerBlocks, RichText } from '@wordpress/block-editor';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { useClientId } from '@crowdsignal/hooks';
+
+/**
+ * Style dependencies
+ */
+import { MultipleChoiceWrapper } from './styles/multiple-choice';
+
+const ALLOWED_ANSWER_BLOCKS = [
+	'core/image',
+	'core/paragraph',
+	'crowdsignal-forms/poll-answer',
+];
+
+const MultipleChoice = ( props ) => {
+	const { attributes, setAttributes } = props;
+
+	useClientId( props );
+
+	const handleChangeQuestion = ( question ) => setAttributes( { question } );
+
+	const handleChangeNote = ( note ) => setAttributes( { note } );
+
+	return (
+		<MultipleChoiceWrapper>
+			<RichText
+				tagName="h3"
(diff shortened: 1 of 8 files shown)


packages/blocks/src/multiple-choice/edit.js
```diff
@@ -0,0 +1,61 @@
...
+	const { attributes, setAttributes } = props;
+
+	useClientId( props );
+
+	const handleChangeQuestion = ( question ) => setAttributes( { question } );
+
+	const handleChangeNote = ( note ) => setAttributes( { note } );
+
+	return (
+		<MultipleChoiceWrapper>
```
ice9js, line 33: Note: in terms of styling, we might want to just create some kind of general `QuestionWrapper` component that's just this wrapper with all the styles.  

I imagine it will/should probably be the same for all question blocks. Added to c/BQqTn3fQ-tr.

packages/blocks/src/multiple-choice/edit.js
```diff
@@ -0,0 +1,61 @@
...
+			<RichText
+				tagName="h3"
+				placeholder={ __( 'Enter your question', 'blocks' ) }
+				onChange={ handleChangeQuestion }
+				value={ attributes.question }
+			/>
+			<RichText
```
ice9js, line 40: Not sure if we should keep this 🤔 
Like, you'd just use a regular paragraph block if you deem it necessary.

ice9js (approved):
LGTM!

packages/blocks/src/multiple-choice/edit.js
CGastrell: Good point, headlines do have a specific use, not sure why I did this.

packages/blocks/src/multiple-choice/edit.js
CGastrell: Ah, didn't think of it this way, but it makes more sense to just wrap the answers. On the other hand, using InnerBlocks prevents us from forcibly assigning a class to those answers. If you look at the style, you'll see I'm using what seems like an autogenerated class name for the answers: https://github.com/Automattic/crowdsignal-ui/pull/44/files/3b16011ff77431c3db0230fb31bb186926f3c712#diff-d9e43fcae2fb4e379078aa8aafb0a3a7d40b0f244b5b965c5da6defb201a9898R8

PR #45, opened 2021-09-17
Refactor BlockEditor component

CGastrell:
As we move towards more control over the editor it makes more sense to remove this thin wrapper (`@crowdsignal/block-editor`) in favor of working directly with the Isolated Block Editor.

This PR unwraps the Isolated Block Editor moving it upwards to the dashboard application/workspace root component.

## Test instructions

Should work as nicely as before


ice9js (approved):
This probably makes the most sense since we decided to go with the 'one editor for all' approach.  

Thanks Christian! :shipit:

PR #46, opened 2021-09-20
Make block name part of the exported settings

ice9js:
This patch updates the blocks' default exports to include the block name as well.  
Reason being, since they need to be 'hardcoded' because of how child-blocks mechanics work, it makes more sense to keep that in one place.

The updated structure also allows for adding any additional metadata to the block if we ever want that.

# Testing

- Run `yarn workspace @crowdsignal/dashboard start`.
- Verify all Crowdsignal blocks still work like they used to.


CGastrell (approved):
LGTM!

PR #47, opened 2021-09-22
Set toolbar to appear above the block instead of being fixed

ice9js:
This is a small patch to address c/tDCumeUy-tr. The toolbar will now appear above the block instead of the main editor toolbar for improved experience.

Screen Shot 2021-09-22 at 4 19 04 PM

# Testing

- Run `yarn workspace @crowdsignal/dashboard start`.
- Check if the toolbar appears on the block instead of the editor toolbar as expected.


digitalwaveride (approved):
Nice! Works as advertised! Thanks Kuba!

PR #48, opened 2021-09-22
Remove left shadow used for hover/active block

CGastrell:
This PR removes the left side shadow for active and hover blocks


digitalwaveride (approved):
🙏 🙏 🙏 🙏 🙏

PR #50, opened 2021-09-24
Fix poll block's number settings not being saved

ice9js:
This patch fixes an issue with numeric attributes being lost on save if they were entered as unparsed numeric strings.

# Testing

- Go to `crowdsignal.localhost:9000/project` and add a poll block.
- Edit the border radius, you should see it change.
- Save draft.
- Refresh the page, the border should still be rounded (previously it'd revert back to square).


CGastrell (approved):
LGTM! Works as advertised :shipit:

PR #51, opened 2021-09-24
Implement form preview

ice9js:
This patch implements a full page preview for created forms which can currently be accessed through `/project/:projectId/preview`.

_Note: The dashboard header is currently included on the page. I opted not to make any changes to our routing just yet until we decide if the preview is actually going to be part of the dashboard or a standalone app._

_Note: As we don't have theme support yet. I temporarily added a minimal wrapper on the page just so we get a similar column as in the editor._

Stuff worth noting:

- The entire preview is rendered as part of the react app using `renderBlocks()`.
- The `/components` directory inside `@crowdsignal/blocks` contains abstractions for the visual parts of the blocks. Ideally, we should avoid adding CSS inside either the editor or the public version of the block unless there's a very specific reason for it.
- I updated the editor blocks to use the said components.
- I've added a set of public-facing versions for our blocks for the preview. This is what we'll later extend with voting logic.

# Testing

- Go to `https://crowdsignal.localhost:9000/project` and add some blocks. Feel free to use a mix of our own and core blocks.
- Go to `https://crowdsignal.localhost:9000/project/xxx/preview` (replace xxx with your project id). You should see a reasonable representation of what you just created.


CGastrell (approved):
Looking good! Could you add:

```js
href={ `/project/${ projectId }/preview` }
```

to the Preview button `apps/dashboard/src/components/editor/toolbar.js:48`?

ice9js:
Good point! Actually, I will be adding the implementation for the whole preview-dropdown menu in a next PR, if that's ok?

CGastrell:
> Good point! Actually, I will be adding the implementation for the whole preview-dropdown menu in a next PR, if that's ok?
:shipit:

PR #53, opened 2021-10-05
Implement form state

ice9js:
This patch adds a new `@crowdsignal/form` package which role is to manage form state when Crowdsignal projects get rendered.

The way it works is the entire form is wrapped with a single `<Form>` component, and any child components can then use `useField` hook to link up to the state.  
The package uses it's own, separate `@wordpress/data` store called `crowdsignal/form` and the form's `name` property is used to store the data separately for each `<Form>` instance.

Due to how the submit mechanics work right now, `<Form>` needs to be an actual `<form>` element but even that could be adapted if necessary - for example should we find ourselves in a scenario where we need to deal with some kind of nested forms.

Other minor changes include:

- I added a `Submit` block which is currently just a button and needs to be added somewhere on the page in order to be able to submit the form.
- I added dedicated `Answer` blocks for `MultipleChoiceQuestionBlock` to replace `PollAnswer` and implemented the correct form mechanics there. (**multi-select buttons work now**)
- I added simple `FormTextInput` and `FormTextarea` components which don't contain too much styling yet but are supposed to be used with the blocks like `FreeText` instead of their `@wordpress/components` or `@crowdsignal/components` counterparts.
- I hooked up `FormTextarea` to the `FreeText` block and linked it up with the form state as well.

# Testing

- Go to `/project` and start a new project.
- Make sure to add free text and multiple choice questions with some answers.
- Add a submit button block.
- Proceed to `/project/{projectId}/preview` using the ID of the project you just created.
- The page should render correctly.
- You should be able to select/deselect MultipleChoice answers - they should change color when you do.
- Hit submit, the current form state will be printed to the console.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -22,6 +22,7 @@
     "@crowdsignal/block-editor": "^0.1.0",
     "@crowdsignal/blocks": "^0.1.0",
     "@crowdsignal/components": "^0.1.0",
+    "@crowdsignal/form": "^0.1.0",
     "@crowdsignal/rest-api": "^0.1.0",
     "@crowdsignal/router": "^0.1.0",
     "@wordpress/blocks": "^11.0.0",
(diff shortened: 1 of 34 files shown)


CGastrell (approved):
Works as described :shipit:

PR #55, opened 2021-10-05
Add toolbar buttons handler/link

CGastrell:
This PR implements 2 simple handlers for our current toolbar so to add functionality on preview and share

* Add a simple link for preview button
* Add a copy-to-clipboard handler for Share button

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`, visit http://crowdsignal.localhost:9000/project

See that, initially, Preview and Share buttons are disabled

  - [ ] make some changes so the autosave function triggers (or hit "Save draft"): "Preview" button should enable
  - [ ] hit "Publish": "Share" button should enable
  - [ ] hit "Share" button: you should get an `alert` modal stating that the URL has been copied to clipboard
  - [ ] confirm the copied URL corresponds to the project

### DIFF
--- a/apps/dashboard/src/components/editor/toolbar.js
+++ b/apps/dashboard/src/components/editor/toolbar.js
@@ -42,12 +42,48 @@ const Toolbar = ( { projectId } ) => {
 			...payload,
 		} );
 	};
+
+	const shareHandler = () => {
+		if ( project.permalink ) {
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Works as advertised! Sweet!

image

PR #56, opened 2021-10-05
Add MC context provider

CGastrell:
This PR wraps MC question block's children in a Context.Provider. The provided value is the attributes from the MC question block.

When the MC allows for multiple selections (not implemented yet) the input naming differs as they are provided in the old PHP form array syntax: `name=input-name[choice][]`. To this end we'll use the MC attributes to determine how the inputs should be named and it would be better to rely on our constants. I left a note to change the (now hardcoded as `0`)  *multiple selection setup* check.

## Test instructions
The changes are not UI noticeable, but the payload printed on the console should look different.

* Checkout and run `yarn workspace @crowdsignal/dashboard start`
* Create/edit a project with a Free Text, an MC and a Submit block
* Publish it and preview it by heading to http://crowdsignal.localhost:9000/project/[ID]/preview (or use the "Preview" button if #55 is already merged)
* Fill and submit

The console should print the values of your answers as an object which keys are strings `q_ + clientId hashed string`.
Each entry on the object would be a key/value pair, FreeText will show your answer as text value while MC will show the clientId of the selected answer.

NOTE: MC response would be an array of selected Answers' clientId hashes once the multiple selection is enabled. 

Something like this:

```
{
    "q_04dc0f6a-245c-4434-af51-83bc3a6f909d[choice]": "03f6ee14-e3b6-4fc4-a28d-51fbda620ee0",
    "q_c5007c0b-a6a4-42e4-bbdc-f1d1c96085df": "free text response"
}
```

@ice9js Let me know if this should be approached differently, for now I just need this to work on the submit payload/process.

### DIFF
--- a/packages/blocks/src/answer/index.js
+++ b/packages/blocks/src/answer/index.js
@@ -3,12 +3,14 @@
  */
 import classnames from 'classnames';
 import styled from '@emotion/styled';
+import { useContext } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import { useField } from '@crowdsignal/form';
 import { Button } from '../components';
+import MultipleChoice from '../multiple-choice';
 
 const Input = styled.input`
 	height: 1px;
@@ -20,8 +22,14 @@ const Input = styled.input`
 const Answer = ( { attributes, className } ) => {
 	const width = attributes.width ? `${ attributes.width }%` : null;
 
+	const parentQuestion = useContext( MultipleChoice.Context );
+
+	// REVIEWER NOTE: choiceType value is assigned from ChoiceType constant under block-editor/multiple-choice/constants,
+	// being 0 = ChoiceType.SINGLE. Hardcoded here as I'm not sure how to import such constant values from the block-editor package.
 	const { inputProps } = useField( {
-		name: `q_${ 'test' }[choice]`,
+		name: `q_${ parentQuestion.clientId }[choice]${
+			parentQuestion.choiceType !== 0 ? '[]' : ''
+		}`,
 		type: 'checkbox',
 		value: attributes.clientId,
 	} );
--- a/packages/blocks/src/multiple-choice/index.js
+++ b/packages/blocks/src/multiple-choice/index.js
@@ -2,11 +2,14 @@
  * External dependencies
  */
 import { RichText } from '@wordpress/block-editor';
+import { createContext } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import { QuestionWrapper } from '../components';
+
+const Context = createContext( 'multiple-choice' );
 
 const MultipleChoice = ( { attributes, children, className } ) => {
 	return (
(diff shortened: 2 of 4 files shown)


packages/blocks/src/answer/index.js
```diff
@@ -20,9 +22,20 @@ const Input = styled.input`
 const Answer = ( { attributes, className } ) => {
 	const width = attributes.width ? `${ attributes.width }%` : null;
 
+	const parentQuestion = useContext( MultipleChoice.Context );
+
+	// REVIEWER NOTE: choiceType value is assigned from ChoiceType constant under block-editor/multiple-choice/constants,
```
ice9js, line 27: `@crowdsignal/block-editor` is dependent on `@crowdsignal/blocks` so whenever dealing with these things, I believe this is the place they should live.  
The idea behind that is: we want to be able to build widget bundles and not include the editor stuff at all.

The `./constants.js` files in `block-editor` are currently kind of a left-over. They were copied there straight from `crowdsignal-forms` just to get the blocks to register in the editor, but now that we have the two packages they belong in `@crowdsignal/blocks`.

To import it inside the block-editor package, I think it's best to make it a property of `MultipleChoice`. So something like this:

```
import { ChoiceType } from `./constants';

const MultipleChoice = ( props ) => {
    ...
};

MultipleChoice.ChoiceType = ChoiceType.

export default MultipleChoice;
```

That said, I don't think there's a need for a `ChoiceType` constant altogether.  
Like we discussed, we need a `minimum/maximum choices` restriction as well. While I'm not sure how exactly we're going to enforce it yet these conditions are absolutely equivalent:

```
(minimumChoices === 1 && maximumChoices === 1) === ChoiceType.SINGLE
```

So I'd just go ahead and use the former and have one less thing to worry about.

packages/blocks/src/multiple-choice/index.js
```diff
@@ -2,21 +2,28 @@
...
  */
 import { RichText } from '@wordpress/block-editor';
+import { createContext } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import { QuestionWrapper } from '../components';
 
+const Context = createContext( 'multiple-choice' );
```
ice9js, line 12: Sidenote: just as a precaution, maybe it'd be a good idea to namespace these with `crowdsignal/...` 🤔

ice9js (approved):
Nice! I left a comment regarding your question but this looks pretty solid 👍

ice9js (changes requested):
I was too quick with the approve thing :( Let's get rid of `ChoiceType` altogether first.

packages/blocks/src/answer/index.js
CGastrell: Thanks! That sounds good.

The ChoiceType thing was derived from our backend classes as a 1:1 correlation, but it's true that we could just rely on max/min and work from there. Mind that, if we go with this "implying" mechanism, a value of `min = 1` would also state the question as mandatory. I think that's the only reason the `type` switch existed: to keep those concepts apart on the UI.

I'll change things and ping you.

PR #57, opened 2021-10-06
Make the content area of the editor scrollable

ice9js:
This patch updates our page and editor layout so that only the editor content is scrollable whenever the user is on the editor page.

Solves c/U2ostitJ-tr.

# Testing

- Go to `/project` and start a new project.
- Add a few block so they no longer fit on the screen
- The editor content should scroll but the header and editor toolbar should stay in place

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -38,6 +38,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


digitalwaveride:
Scrolling kind of works... but Firefox gives me not much space ;) 

https://user-images.githubusercontent.com/5120357/136196098-a637a23c-59a7-413e-8fbc-30ac4a58c65a.mov

image

PR #58, opened 2021-10-06
Make project title editable

ice9js:
This patch solves c/NXdi1uep-tr and enables editing of the project title.

While working for this I had a few concerns due to how we currently implement the project state:

- As or right now, we have no complete reference for the `project` object anywhere in our codebase which makes it slightly difficult to work with. At least it'd be nice to explicitly list all the properties inside the reducer by using something like `pick` or even implement some proper validation for it.
- Because of how `project` is stored, and especially because of how `saveAndUpdateProject` works, any component that updates any of its properties is now closely coupled to the actual `project` structure. To remedy this we can either update our project state setup and provide more granular control for the saving side-effect, or at the very least, we should provide a set of helper functions for working with a `project` so you'd call stuff like `project = updateProjectTitle( project, title );` instead of modifying it directly.

# Testing

- Go to `/project` and create a project, it'll have `Drafted!` as the default title.
- Click on the title to edit it.
- After making the changes, click anywhere outside of the title to stop editing.
- The changes should be saved immediately afterwards.
- Reload the page and verify the changes have been saved.

### DIFF
--- a/apps/dashboard/src/components/editor/toolbar.js
+++ b/apps/dashboard/src/components/editor/toolbar.js
@@ -22,15 +22,13 @@ const Toolbar = ( { projectId } ) => {
 	} );
 
 	const syncProject = () => {
-		const payload = { title: 'Drafted!' };
 		saveAndUpdateProject( projectId, {
 			...project,
-			...payload,
 		} );
 	};
 
 	const publishProject = () => {
-		const payload = { title: 'Published!', publish: true };
+		const payload = { publish: true };
 		saveAndUpdateProject( projectId, {
 			...project,
 			content: {
--- a/apps/dashboard/src/components/project-navigation/index.js
+++ b/apps/dashboard/src/components/project-navigation/index.js
@@ -1,12 +1,14 @@
 /**
  * External dependencies
  */
+import { useDispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
-import { PageHeader, TabNavigation } from '@crowdsignal/components';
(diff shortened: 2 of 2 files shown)


CGastrell (approved):
Looks good, works as advertised!

Maybe it would be nice to remove the hardcoded stuff when saving draft and publishing:

https://github.com/Automattic/crowdsignal-ui/blob/4346aa8a2f1eb164db1597bb854bc013c560e07e/apps/dashboard/src/components/editor/toolbar.js#L25

https://github.com/Automattic/crowdsignal-ui/blob/4346aa8a2f1eb164db1597bb854bc013c560e07e/apps/dashboard/src/components/editor/toolbar.js#L33

Which makes me think on those "race conditions" you were talking about as the Draft/Publish mechanisms don't know anything about the title. What about an action that just updates the project title on the store, but triggers no request until auto-save or Draft/Publish is hit?

I'm giving a thumbs up in case you prefer to address that in a separate change, but if you feel like doing it here, I'll review again

ice9js:
>Which makes me think on those "race conditions" you were talking about as the Draft/Publish mechanisms don't know anything about the title. What about an action that just updates the project title on the store, but triggers no request until auto-save or Draft/Publish is hit?

So this is a problem because the 'autosave' is currently part of the editor instead of a separate layer or something that runs in the background. As such, there's a chance it'll never get `saved` after you update the title.  
My main concern is encapsulating the project related logic since it's a large object and especially working with drafts, published versions and pages could get messy really quickly. We need to figure out what direction to take with that and then we'll just deal with saving appropriately afterwards.

Anyway, thanks for the reminder on the hardcoded titles - I forgot that was there.

PR #61, opened 2021-10-11
Update project renderer app

CGastrell:
This PR updates the `project-render` app to work on and debug.

The purpose of this update is to build a working version for D68245-code

NOTE: A build automation should make for a more easy-to-update build and deploy process

## Test instructions
As the full async process is not ready yet, a hardcoded project layout is provided. Checkout and run `yarn workspace @crowdsignal/project-renderer start`. Go t http://crowdsignal.localhost:9000. You should see a simple project with a Free Text and a Multiple Choice questions, and a Submit button.

If the mentioned differential has been applied, you should be able to submit. Otherwise, a CORS/NOT FOUND error should occur. It might be best to try it out on the differential as the build has already been put into place there.


ice9js (approved):
Code looks good. I do have one recommendation going forward though - while we might still rely on `window.__CS_PROJECT__` for now, it might make sense to refactor and organize our code as if we were fetching it asynchronously already.  

Incorporating those early might save us a few headaches later down the line when there's a ton of stuff going on.  
Either way, that can come in a separate PR.

PR #63, opened 2021-10-13
Add project state button behavior

CGastrell:
This PR adds the necessary state/reducer/action/selector combo so the "Save Draft", "Share" and "Publish/Update" buttons reflect the state of the Project.

`changeProject` action: triggers `PROJECT_CHANGE` with no extra payload for now. Used to reduce to `false` both `isDraftSaved` and `isPublishSaved` so the buttons can enable/disable accordingly

The contents now also carry a timestamp from backend, so this PR should be tested against D68353-code 

NOTE: we could use `select( 'core/block-editor' ).isTyping()` instead/on top of the debounced save for better UX?

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`. Use pabtAt-1md-p2 as a guide on button behavior.

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -58,7 +58,9 @@ const Editor = ( { projectId } ) => {
 		draftedBlocks
 	);
 
-	const { saveAndUpdateProject, changeProject } = useDispatch( STORE_NAME );
+	const { saveAndUpdateProject, changeProjectContent } = useDispatch(
+		STORE_NAME
+	);
 
 	const debounceSave = useCallback(
 		debounce( ( content ) => {
@@ -92,7 +94,7 @@ const Editor = ( { projectId } ) => {
 	const handleChangeContent = useCallback(
 		( content ) => {
 			if ( isSaved ) {
-				changeProject( project );
+				changeProjectContent( project );
 			}
 			debounceSave( content );
 		},
--- a/apps/dashboard/src/data/projects/actions.js
+++ b/apps/dashboard/src/data/projects/actions.js
@@ -15,7 +15,7 @@ import {
 } from '@crowdsignal/rest-api';
 import { dispatchAsync } from '../actions';
 
-export function changeProject( project ) {
+export function changeProjectContent( project ) {
 	return {
 		type: PROJECT_CONTENT_CHANGE,
 		project,


apps/dashboard/src/data/action-types.js
```diff
@@ -19,6 +19,8 @@ export const POLL_UPDATE = 'POLL_UPDATE';
 export const PROJECT_SAVE = 'PROJECT_SAVE';
 export const PROJECT_UPDATE = 'PROJECT_UPDATE';
 export const PROJECT_SAVE_ERROR = 'PROJECT_SAVE_ERROR';
+export const PROJECT_LOADED = 'PROJECT_LOADED';
```
ice9js, line 22: This doesn't seem to be used anywhere. Let's remove it.

apps/dashboard/src/data/action-types.js
```diff
@@ -19,6 +19,8 @@ export const POLL_UPDATE = 'POLL_UPDATE';
...
+export const PROJECT_CHANGE = 'PROJECT_CHANGE';
```
ice9js, line 23: This is a little less descriptive than I'd have hoped, can we change it to `PROJECT_STATUS_UPDATE` ?

We can then create an 'enum' for the different status values as required by the frontend, so in addition of updating the actual project object, these could be stored under a separate reducer key.

As for where the enum should go: `/src/util/project/status.js` - we already have some in `/src//util/user/...` if you need a reference.

Suggested values:

`DRAFT`
`DRAFT_WITH_UNSAVED_CHANGES`
`PUBLIC`
`PUBLIC_WITH_UNSAVED_CHANGES`
`PUBLIC_WITH_UNPUBLISHED_CHANGES`

A bit lengthy but I think it gives the right picture. If anything they'll be easy to rename once we start using them.

apps/dashboard/src/data/projects/reducer.js
```diff
@@ -46,8 +48,44 @@ const isSaving = ( state = false, action ) => {
 	return state;
 };
 
+const isDraftSaved = ( state = false, action ) => {
```
ice9js, line 51: Like mentioned before, I believe both `isDraftSaved` and `isPublishSaved` could be replaced by a single `status` reducer.  

Note: I think if we go this way this one should best be 'keyed', similar to `items`.

ice9js (changes requested):
Overall, this seems to be the right direction but I left some comments on how to possibly make the code a little more concise.

apps/dashboard/src/data/projects/reducer.js
```diff
@@ -46,8 +49,41 @@ const isSaving = ( state = false, action ) => {
...
+	const hasBeenPublished = get(
+		action,
+		[ 'project', 'content', 'published' ],
+		null
+	);
+	if ( action.type === PROJECT_UPDATE ) {
+		newStatus = ProjectStatus.DRAFT;
+		if (
+			hasBeenPublished &&
+			action.project.content.published.ts <
```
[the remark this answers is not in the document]
CGastrell: I'll see to that on the backend

.eslintrc
```diff
@@ -3,7 +3,7 @@
 	"rules": {
 		"complexity": [
 			"error",
-			10
+			11
```
ice9js, line 6: 😆

PR #69, opened 2021-10-20
Fix renderBlocks' parseNodes function

ice9js:
This patch fixes issues with some core blocks causing our preview/page renderer not to work correctly or crash altogether.  
The main culprits I found were:

- Usage of lodash' `lowerCase` instead of `toLower` causing tag names to be converted in a wrong way.
- We didn't support for any HTML attributes on these blocks other than `class`, thus blocks like the image block would never work.

Solves c/2OdLDU86-tr.

# Testing

- Create a new project and proceed to add core blocks to it. Header and Image blocks in particular as they've shown to be problematic in the past.
- Go to the preview.
- The page should now display correctly.

### DIFF
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -9,9 +9,10 @@ import {
 	isEmpty,
 	isString,
 	keys,
-	lowerCase,
+	toLower,
 	map,
 	uniqueId,
+	zipObject,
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Fixed! Basic core blocks are working! 😻 
Thanks Kuba!

PR #74, opened 2021-10-28
Simplify block loader

ice9js:
While working on #73, I noticed the `<BlockLoader />` component does very little compared to using isolated block editor's own `onLoad` callback.  

Unfortunately `updateBlockWithoutUndo` doesn't work in either case and looks like it's an issue that needs to be addressed upstream. As such, I'm not sure if there's any other reason we should hold onto our custom loader?

# Testing

Open an existing project in the editor and verify it's still loading correctly as expected.


CGastrell (approved):
LGTM!

PR #75, opened 2021-10-29
Add type checking capabilities to help deal with convoluted data structures

ice9js:
This is an initial stab at trying to make parts of our application type-safe.

Solves c/MV5belpY-tr.

We have a few objects that can get quite large or chaotic, think of project or user state. I'd like to organize them a little better by defining them in proper types as well as provide some helpers to deal with updating any of these objects. This way we could keep all related logic in one place instead of manipulating objects manually inline - resulting in duplicated, more error prone code that quickly becomes difficult to understand as necessary context is spread all over the place.

As part of this endeavour, I also implemented a type-safe `curry()` function so our helpers are compatible with `flow`,`flowRight`,`compose` and the like, all while maintaining type safety.

That said, it's not all rainbows and butterflies so here's a quick overview of my findings so far regarding strong typing:

### The good

- Types. We now have a good reference of what data structures we're working with.
- Safety. When type checking is enabled and proper types are inferred or specified explicitly, you will get an error if you're trying to access a property that doesn't exist. Similarly we can ensure certain structures must exist in certain objects, saving us a null-check here and there.
- Syntax. With typescript's JSDoc syntax, we don't have to do much more than remember to add our regular docBlocks above our functions and it automatically picks it up.
- We don't need to update our toolchain - it's still all JS.

### The bad

- As not everything we write is fully typed, Typescript cannot fully deduce types where it otherwise should and requires explicit `@type` definitions.
- Somewhat related to the above, as our type definitions sit in their own package, importing them is quite painful. For example: `@typedef {import("../../../../../packages/types/src/project").Project} Project`. It'd be helpful if we can make them available globally across our modules.
- We can't restrict type-checking to certain types only. As such, whenever it's enabled for a file, the entire file would need to be updated to fix any errors that might pop up - including ones unrelated to the current issue.

### Possible compromise

If we forget about types for a minute, I think the helpers themselves are great. After all:

```javascript
const publicProject = publishProject( project );
```

Reads much better than:

```javascript
const publicProject = {
    ...project,
    content: {
        ...project.content,
        public: project.content.draft,
    },
    public: true,
};
```

The two new packages I added are fully strongly-typed. That already ensures some safety between the helpers themselves. Moving that logic into dedicated functions also makes them testable. I think that's definitely a part worth keeping.

As for enforcing types across the rest of our app... The main benefit would be the added safety when accessing any object properties. The overhead is not much but it's there.  
It's worth noting the same level of safety could be achieved by implementing similar helpers for accessing properties rather than just updating our big objects. This also has the added benefit of totally encapsulating the internal composition of any such object.  
This is the part I think needs some discussion.

### Todo

- [ ] Add a doc describing the overall stance on types across the entire repo.
- [ ] Convert project's returned from the API into the structure used in the app and vice-versa.

# Testing

Run `yarn check-types` to validate there aren't any type related errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,8 @@
 			"error",
 			10
 		],
+		"jsdoc/require-property-description": "off",
+		"jsdoc/require-returns-description": "off",
 		"jsx-a11y/anchor-has-content": [
 			"off"
 		],
@@ -37,6 +39,9 @@
 			"node": {
 				"paths": ["apps/dashboard/src"]
 			}
+		},
+		"jsdoc": {
+			"mode": "typescript"
 		}
 	}
 }
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -26,6 +26,7 @@
     "@crowdsignal/hooks": "^0.1.0",
     "@crowdsignal/rest-api": "^0.1.0",
     "@crowdsignal/router": "^0.1.0",
+    "@crowdsignal/types": "^0.1.0",
     "@wordpress/blocks": "^11.0.0",
     "@wordpress/data": "^6.0.0",
     "@wordpress/element": "^4.0.0",
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -1,17 +1,20 @@
+// @ts-check
+
 /**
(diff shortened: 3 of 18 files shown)


ice9js:
Superseded by #132.

PR #64, opened 2021-10-13
Add editor sidebars

ice9js:
This PR enables editor sidebars for the inserter, list view, and block and document settings.  

crowdsignal localhost_9000_project_2670174 (3)

Solves c/zJuK6WL8-tr, c/gJKCWpqO-tr and c/mmjwQUjE-tr.
Depends on [Automattic/isolated-block-editor#75](https://github.com/Automattic/isolated-block-editor/pull/75/files).

To do:
- Update isolated-block-editor to the correct version once the above PR gets merged.
- It'd be great if we can manually force the inserter open, and perhaps force it to stay open even when not in focus.

Needs #94 to be merged first.

# Testing

- Go to `/project(/projectId)` to open or start a project
- Clicking on the `+` button in the upper left corner should open the inserter in a sidebar on the left.
- There should be a 'list view' icon (4th from the left). Clicking on it will display the list view of the project in a sidebar on the left.
- There shouldn't be an 'info' icon on the left side of the editor toolbar.
- Clicking the cog icon in the upper right should open the document and block settings in a sidebar on the right.

PR #77, opened 2021-11-01
Update Automattic/isolated-block-editor to version 2.6.0

ice9js:
This patch updates our Automattic/isolated-block-editor dependency to 2.6.0 (latest) that includes the latest Gutenberg.  

Upgrading our editor also enables a bunch of fixes from #64 to be merged as well as it looks it'll still be a while before the sidebars get merged. It's also required in order to complete #76.

# Testing

Run `yarn install` and `yarn workspace @crowdsignal/dashboard start` to verify the editor still runs as expected after the upgrade.


CGastrell (approved):
Works fine! :shipit:

PR #79, opened 2021-11-04
Disable 'publish' button when the page doesn't contain any submit buttons

ice9js:
This patch will disable the _publish_ button if the current page doesn't contain a submit button.  
When it's disabled, users can trigger a popover giving the necessary context by hovering over the button.

Solves c/aev5y2xb-tr.

### DIFF
--- a/apps/dashboard/src/components/editor/toolbar.js
+++ b/apps/dashboard/src/components/editor/toolbar.js
@@ -2,15 +2,15 @@
  * External dependencies
  */
 import { Button } from '@wordpress/components';
+import { useDispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 import { ToolbarSlot } from 'isolated-block-editor'; // eslint-disable-line import/named
 
 /**
  * Internal dependencies
  */
+import PublishButton from './publish-button';
 import { STORE_NAME } from '../../data';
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/components/editor/toolbar.js
```diff
@@ -74,46 +65,37 @@ const Toolbar = ( { projectId } ) => {
 
 	return (
 		<ToolbarSlot className="block-editor__crowdsignal-toolbar">
-			<Button
-				className="is-crowdsignal"
+			<ToolbarButton
```
CGastrell, line 68: I think I missed something here... so, creating a style and then using it with `as={ Component }` is enough? We don't need to "wrap" the component with a styled component? Sounds good, it's just that I never seen it like this before

CGastrell (approved):
Nice! :shipit:

apps/dashboard/src/components/editor/toolbar.js
ice9js: Yeah. It's quite useful - we actually use it like that a lot in our block-editor package where we want `RichText` etc. be styled as well.

There's also a `cx` helper in `@emotion` which kind of acts like `classnames` for these. But in general, such cases would be more about the styled component being defined as a certain element, and us wanting to render it as a different one or as a React component with just the same styles applied to it.

PR #80, opened 2021-11-04
Remove a duplicated import from the Editor component

ice9js:
Small patch to remove a duplicated import on our `<Editor />` component which is a result of a recent rebase and is causing `yarn lint:js` to throw errors.  

# Testing

Run `yarn lint:js` and verify it doesn't throw any errors.


CGastrell (approved):
LGTM!

PR #81, opened 2021-11-04
Update dev server port for project renderer to avoid collision with the dashboard build

ice9js:
This patch updates the dev server port for the project renderer from `9000` to `9001` so it doesn't clash with the dashboard build and the two can potentially be run at the same time.

# Testing

- Run `yarn workspace @crowdsignal/project-renderer start` and wait for the project to build.
- The project renderer should now be available under `https://crowdsignal.localhost:9001`.


CGastrell (approved):
:shipit:

PR #82, opened 2021-11-04
Add a simple 'Thank You' page for the project renderer

ice9js:
This patch adds a most simplistic 'thank you' page that'll be displayed after a respondent has successfully submitted their response.  
Currently, the page consists of just a headline stating `Thank you for your response!`.

# Testing

- Run both `yarn workspace @crowdsignal/dashboard start` and `yarn workspace @crowdsignal/project-renderer start`
- Create a new project, add at least one question and a submit button, save and publish it.
- Visit the project using the project renderer. When you correctly submit a response, the thank you message should be shown.

_Note: Project renderer works with project codes rather than IDs. Use the `share` button to get the project code and insert it into the local project renderer url: `https://crowdsignal.localhost:9001/{projectCode}`._

### DIFF
--- a/apps/project-renderer/package.json
+++ b/apps/project-renderer/package.json
@@ -25,6 +25,7 @@
     "@crowdsignal/hooks": "^0.1.0",
     "@crowdsignal/rest-api": "^0.1.0",
     "@wordpress/element": "^4.0.2",
+    "@wordpress/i18n": "^4.2.3",
     "lodash": "^4.17.21"
   },
   "devDependencies": {
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -1,9 +1,10 @@
 /**
  * External dependencies
  */
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
 import styled from '@emotion/styled';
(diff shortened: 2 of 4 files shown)


digitalwaveride (approved):
Nice Kuba!!!

Submit page worked as advertised!

There were three little things I noticed along the way... 
... likely not related to this PR just to mention it here in case... 
If not related, could you check please for which of these below we should create a new Trello card in case of not a known issue? Thank you! 🙏 

#### Submit Button Styling

Some issues with the button width... 

image


#### Project renderer styling

Styling for the project renderer seems to be off compared to editor & preview

Here a side by side ... editor | preview | renderer

image


#### Results Page

Still super excited when seeing responses ending up on the results page. One thing I noticed when going to the results page is that the "Title of the Project" is behaving strangely. When opening the results the title was active in edit mode, when I clicked away it gave me the can't be empty error ... 

It seemed as it would not have stored the value for the title that it had before ... I guess it was "untitled survey".

image

Then when I try to enter a title it does not save it and I am staying stuck with the error.

Also, for the survey page the Subnavigation with Settings, Style etc. is still showing.

image

CGastrell (approved):
Simple, yet effective :shipit:

PR #83, opened 2021-11-05
Fix style attribute parsing for renderBlocks()

ice9js:
This patch fixes how `style` attributes are parsed in `renderBlocks()`.

When we parse block HTML, they're read as strings but React expects `style` to be an object causing the page to crash. We need to manually convert the string into an object for it to work.

# Testing

- Create a new project and add blocks that rely on the style attribute being saved in their HTML output. For example `Columns` and `Spacer` blocks.
- Verify these blocks don't crash the preview and have the expected width/height as set in the editor.

### DIFF
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -3,6 +3,7 @@
  */
 import { createElement } from '@wordpress/element';
 import {
+	camelCase,
 	filter,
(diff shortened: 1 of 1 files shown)


digitalwaveride (approved):
Thanks for fixing Kuba!!! 

🚀 Works!

image

PR #84, opened 2021-11-08
Enable block wide alignment

ice9js:
This patch enables wide alignment inside of our editor. Part of c/QsghLr4q-tr.

# Testing

- Add a block that supports wide alignment. For example an image block.
- The alignment setting in the toolbar should now show additional "Wide" and "Full width" settings.
- Selecting the settings should be reflected in block's appearance - the image should now exceed the width of the content column.

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -37,6 +37,10 @@ const editorSettings = {
 			toc: false,
 		},
 	},
+	editor: {
+		alignWide: true,
+		supportsLayout: false,
+	},
 };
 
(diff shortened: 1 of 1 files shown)


digitalwaveride:
New width options work in the editor ... 

image

... seems not to be applied though yet to the preview and project renderer?

https://crowdsignal.localhost:9000/project/2688007/preview

CGastrell (approved):
LGTM!

ice9js:
Yeah, in terms of the preview, it seems the main theme styles don't contain the rules for block alignment.  
We'll also need to update our content wrapper. Reported in c/H3p0CZCg-tr.

PR #85, opened 2021-11-09
Offer to restore unpublished changes (onload)

CGastrell:
This PR adds a blunt window.confirm to restore unpublished changes upon loading a project.

The solution is a rough one, albeit effective, but it will serve until we put in place some more fancy prompt.

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`. Visit http://crowdsignal.localhost:9000/project and add some content. Publish it.
Make some more changes, let it autosave or hit "Save draft" so you now have a newer draft version.
Reload the page. When the page loads, it will prompt you to restore the unpublished changes.
Hit "Ok" to restore. Hit "Cancel" to load the published version on the editor.
Reload as many times to check what happens by clicking "Ok"/"Cancel". Finally, hit "Update" button and reload again, the prompt should not show


ice9js (approved):
Works like described 👍

PR #86, opened 2021-11-09
Disable non-working core blocks

ice9js:
This patch disables some core blocks that currently don't work in our editor, to the point where it doesn't make sense to even keep them as an option. The complete list:

- core/audio
- core/calendar
- core/cover
- core/embed
- core/file
- core/media-text
- core/shortcode

Part of c/QsghLr4q-tr.

# Testing

Try adding any of the above blocks. They shouldn't be available in the inserter anymore.

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -1,12 +1,11 @@
 /**
  * External dependencies
  */
+import { parse } from '@wordpress/blocks';
 import { useCallback } from '@wordpress/element';
 import { useDispatch, useSelect } from '@wordpress/data';
-import { __ } from '@wordpress/i18n';
 import { debounce, get, noop } from 'lodash';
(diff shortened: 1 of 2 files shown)


digitalwaveride:
Any chance to "recover" the "Cover block" ? ;) 

Making it work similar to the image block? Pasting URLs for showing images... 
It would be a very interesting block to create content with background!

ice9js:
>Any chance to "recover" the "Cover block" ? ;)
>Making it work similar to the image block? Pasting URLs for showing images...

Unfortunately not. It's a core block which very much makes this a take it or leave it situation.
In order to do what you're suggesting, we'd need to create our own cover block.

CGastrell (approved):
:( LGTM

digitalwaveride:
> Unfortunately not. It's a core block which very much makes this a take it or leave it situation.

Gotcha! Let's wait on our medial library integration then 👍

PR #87, opened 2021-11-10
Add a "Questions" block category for crowdsignal question blocks

ice9js:
This patch adds a dedicated 'Questions' category for our question blocks and positions it on top of the inserter in the editor.

Solves c/4StsZCS0-tr.

Screen Shot 2021-11-10 at 2 48 28 PM

# Testing

Open the inserter in the editor - it should now feature the new 'Questions' category at the top with, for now, our submit button, multiple choice and text question blocks.


CGastrell (approved):
Is the `crowdsignal-forms/questions` slug/category portable to Gutenberg? I mean, I guess at some point we'll want to (via plugin or something) make these blocks available on .com

Anyways, LGTM! :shipit:

ice9js:
>Is the crowdsignal-forms/questions slug/category portable to Gutenberg? I mean, I guess at some point we'll want to (via plugin or something) make these blocks available on .com

Yeah. Each block is responsible for specifying the category slug it belongs to and as for categories themselves... WordPress allows it but actually [handles it in the backend](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#block_categories_all).

digitalwaveride:
Thank you Kuba! 🙏 

Two little things:

#### Name of the category

"Questions" feels like it makes sense now, but soon a bunch of form blocks are coming up, and the "submit button" feels a tiny bit weird in that category as well.

The overarching category for questions, forms, as well as things like "submit buttons", maybe in future also something like "results blocks" etc. would be in the end "form". All is part of a "form" ;) 

I would suggest to switch the category to "Form" for now.


#### Icons of blocks

While we have our eyes on this, could we please update the icons of these blocks?
SVGs are shared in the [Trello card.](https://trello.com/c/aIW4JqGx)

image

CGastrell:
What about "Inputs" as category?

digitalwaveride:
Well... we can call them "Blocks"... as that's what they all... ;) 

"Inputs", as technically not wrong, but it feels like a "seize too big" ... I think. 

All the blocks in that category later on will also likely live in WP Gutenberg inside that big wrapper block "form".

CGastrell:
Since this is our own "universe" of blocks, we could also just name the category "Crowdsignal", "Form elements"... or whatever we need it to be so it becomes clear to the user what those are :)

ice9js:
I updated the name and added the icons we have so far - although some are still missing (submit button, multiple choice answer) but can be added at a later time.

And just for the record, the category label or name isn't set in stone. We can easily rename it or split it up later.

PR #88, opened 2021-11-10
Remove the Poll block

CGastrell:
This PR removes the registration and initialization of the Poll block.

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`. Go to crowdsignal.localhost:9000/project and attempt to add a Poll block, you shouldn't be able to do so.

@ice9js any other cleanup we need to do here? We might do good and get rid of the whole store/data in place, but not sure if you had plans for it


digitalwaveride (approved):
It's gone! 👍 

image

PR #90, opened 2021-11-12
Fix MC answer blocks

CGastrell:
This PR addresses a glitch happening on the MultipleChoiceAnswer blocks.

The glitch was introduced due to a condition/assignment typo:

```
const styleClass = className && className.match( /is-style-([^\s]+)/i )[ 1 ];
// changed to:
const styleClass = className && className.match( /is-style-([^\s]+)/i );
```

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`, create a project at http://crowdsignal.localhost:9000/project 
Add a MultipleChoice Question block.

- [ ] see that the answer options look ok

Add a couple of answer options (MultipleChoice Answer) and publish it. Copy the public URL by clicking on the "Share". Open a new tab, paste the URL and append `?use-project` 

- [ ] you should see the public version of the project
- [ ] MC answers should look ok

### DIFF
--- a/packages/blocks/src/util.js
+++ b/packages/blocks/src/util.js
@@ -5,8 +5,6 @@
  * @return {string}           Block style
  */
 export const getBlockStyle = ( className ) => {
-	const styleClass =
-		className && className.match( /is-style-([^\s]+)/i )[ 1 ];
-
+	const styleClass = className && className.match( /is-style-([^\s]+)/i );
 	return styleClass ? styleClass[ 1 ] : '';
 };


digitalwaveride:
Thanks Christian for looking into this! 🙏 


I was testing this as always on FF: 

### Editor
Answer options are back. 👍👍👍

The styling of the buttons is weird, but that is a different issue I guess?

image

### Preview
Looks nearly like the editor, submit button styling is different though... 

image


### Renderer
In the renderer the styling of the answers is a bit more extreme 🤷‍♂️ 

image

CGastrell:
Weird, this is what I'm seeing in the renderer:
image

ice9js (approved):
Thanks Christian!

As for button styles, especially the submit button, it seems that's a regression that was accidentally introduced in one of the other PRs. I think it was #82.  
Anyway, let's get this merged and we'll fix those separately.

PR #92, opened 2021-11-23
Create TextInput block

jcheringer:
## Description

This PR is to create a new block (Text Input) that will be used, mainly, to create forms like contact, address, etc.

Ref: c/pTOrFn0d-tr

### Reference Image
image

## Test Instructions

* Go to the project editor (`project/:projectId`)
* Click the plus button to add a new block and search for `Text Input`, or type `/text input`
* You should be able to add the new block to the page and play with its settings

### DIFF
--- a/packages/block-editor/src/text-input/attributes.js
+++ b/packages/block-editor/src/text-input/attributes.js
@@ -7,25 +7,11 @@ export default {
 		type: 'boolean',
 		default: false,
 	},
-	labelColor: {
-		type: 'string',
-	},
 	textColor: {
 		type: 'string',
 	},
 	backgroundColor: {
 		type: 'string',
-	},
-	borderColor: {
-		type: 'string',
-	},
-	borderWidth: {
-		type: 'number',
-		default: 2,
-	},
-	borderRadius: {
-		type: 'number',
-		default: 0,
 	},
 	inputHeight: {
 		type: 'number',
--- a/packages/block-editor/src/text-input/attributes.js
+++ b/packages/block-editor/src/text-input/attributes.js
@@ -15,10 +15,10 @@ export default {
 	},
 	inputHeight: {
 		type: 'number',
-		default: 30,
+		default: 40,
 	},
-	width: {
-		type: 'number',
-		default: 100,
+	inputWidth: {
+		type: 'string',
+		default: '100%',
 	},
 };
--- a/packages/block-editor/src/text-input/attributes.js
+++ b/packages/block-editor/src/text-input/attributes.js
@@ -1,0 +1,38 @@
+export default {
+	label: {
+		type: 'string',
+		default: '',
+	},
+	mandatory: {
+		type: 'boolean',
+		default: false,
+	},
+	labelColor: {
+		type: 'string',
+	},
+	textColor: {
+		type: 'string',
+	},
+	backgroundColor: {
+		type: 'string',
+	},
+	borderColor: {
+		type: 'string',
+	},
+	borderWidth: {
+		type: 'number',
+		default: 2,
+	},
+	borderRadius: {
+		type: 'number',
+		default: 0,
+	},
+	inputHeight: {
+		type: 'number',
+		default: 30,
+	},
+	width: {
+		type: 'number',
+		default: 100,
+	},
+};
--- a/packages/block-editor/src/text-input/edit.js
+++ b/packages/block-editor/src/text-input/edit.js
@@ -24,7 +24,7 @@ const EditTextInput = ( props ) => {
 				onChange={ handleChangeLabel }
 				value={ attributes.label }
 			/>
-			<FormTextInput
+			<FormTextInput.Preview
 				style={ {
 					width: `${ attributes.width }%`,
 					height: `${ attributes.inputHeight }px`,
--- a/packages/block-editor/src/text-input/edit.js
+++ b/packages/block-editor/src/text-input/edit.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { RichText } from '@wordpress/block-editor';
+import { ResizableBox } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 
 /**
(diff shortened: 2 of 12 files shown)


packages/block-editor/src/text-input/sidebar.js
```diff
@@ -0,0 +1,66 @@
+/**
+ * External dependencies
+ */
+import {
+	// eslint-disable-next-line
+	__experimentalToggleGroupControl as ToggleGroupControl,
+	// eslint-disable-next-line
+	__experimentalToggleGroupControlOption as ToggleGroupControlOption,
```
jcheringer, line 8: Those two confused me a bit. Checking the `@wordpress/components` storybook, I could see that `ToggleGroupControl` and `ToggleGroupControlOption` are not under the experimental section. Should we update our `@wordpress/components` version?

packages/block-editor/src/text-input/sidebar.js
```diff
@@ -0,0 +1,66 @@
...
+				<ToggleControl
+					label={ __( 'The field is required' ) }
+					checked={ attributes.mandatory }
+					onChange={ handleChangeAttribute( 'mandatory' ) }
+				/>
+				<TextControl
+					label="Input field height"
+					type="number"
+					value={ attributes.inputHeight }
+					onChange={ handleChangeAttribute( 'inputHeight' ) }
```
ice9js, line 44: I believe there might be issue with this one.  
`handleChangeAttribute` saves the attribute as is, which means it'd attempt saving `inputHeight` as a string which will work in the editor but the changes will not actually get persisted when you publish or refresh the page.  
To solve this we need another helper and run `parseInt` on the value before it's passed to `setAttribtues`.

packages/block-editor/src/text-input/sidebar.js
```diff
@@ -0,0 +1,66 @@
...
+				/>
+				<ToggleGroupControl
+					onChange={ handleChangeAttribute( 'width' ) }
+					value={ attributes.width }
+					label={ __( 'Field Width' ) }
```
ice9js, line 49: The translation domain is missing here.

packages/icons/src/text-input.js
```diff
@@ -1,4 +1,4 @@
-export const TextInput = () => (
+export const TextInputIcon = () => (
```
ice9js, line 1: 👍

packages/block-editor/src/text-input/edit.js
```diff
@@ -0,0 +1,37 @@
...
+const TextInput = ( props ) => {
+	const { attributes, setAttributes } = props;
+
+	const handleChangeLabel = ( label ) => setAttributes( { label } );
+
+	return (
+		<div style={ { ...useColorStyles( attributes ) } }>
+			<Sidebar { ...props } />
+			<RichText
+				placeholder={ __( 'Enter form label', 'blocks' ) }
```
ice9js, line 23: Translation domain should follow the package name.

ice9js (changes requested):
Looking good! I added a few minor comments and it should be good to go.

packages/block-editor/src/text-input/index.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+import { TextInputIcon } from '@crowdsignal/icons';
+import attributes from './attributes';
+import Edit from './edit';
+
+const name = 'crowdsignal-forms/text-input';
+
+const settings = {
+	apiVersion: 1,
+	title: __( 'Text Input Form', 'blocks' ),
```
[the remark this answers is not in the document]
jcheringer: Got it! Thanks for explaining it.

packages/block-editor/src/text-input/edit.js
```diff
@@ -0,0 +1,37 @@
...
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { FormTextInput } from '@crowdsignal/blocks';
+import { useColorStyles } from '@crowdsignal/styles';
+import Sidebar from './sidebar';
+
+const TextInput = ( props ) => {
```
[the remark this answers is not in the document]
jcheringer: I agree, sounds good. ✅

packages/block-editor/src/text-input/attributes.js
```diff
@@ -0,0 +1,38 @@
+export default {
+	label: {
+		type: 'string',
+		default: '',
+	},
+	mandatory: {
+		type: 'boolean',
+		default: false,
+	},
+	labelColor: {
```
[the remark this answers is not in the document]
jcheringer: Yes! Forgot to remove this. 😅

packages/block-editor/src/text-input/sidebar.js
jcheringer: Nice catch! Thanks for clarifying.

packages/block-editor/src/text-input/sidebar.js
jcheringer: Actually, I added those two. Although they are not under the experimental section on Storybook, even in the latest version, the packages are exported as experimental:

https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/index.js#L138

ice9js: Let's keep it then 👍

ice9js (changes requested):
Looking great! I've got just one more request:

Inside the editor, you shouldn't be able to edit the content of the input itself - similar to how `TextQuestion` works.  
Instead of rendering it as disabled - which might alter its styles - we can take a similar approach to `FormTextarea`. Which is to define a `FormTextInput.Preview` component which contains an overlay that prevents any interaction with the input itself.

jcheringer:
> Looking great! I've got just one more request:
> 
> Inside the editor, you shouldn't be able to edit the content of the input itself - similar to how `TextQuestion` works. Instead of rendering it as disabled - which might alter its styles - we can take a similar approach to `FormTextarea`. Which is to define a `FormTextInput.Preview` component which contains an overlay that prevents any interaction with the input itself.

Yeah, that makes sense. Just did it. 😃

digitalwaveride:
Great progress João! First block is coming together 🥳 
The spacing between the blocks when added to a columns set up feels not bad. 👍 


A few things that are still missing: 

#### Toolbar features
image

- The alignment options
- Mandatory toggle

#### Input Field Resizing Feature

That includes the sidebar integration... 
image

... and the "drag" handlers on the input field.
image

#### Input Field Styling

Height of the field should be 40px.
image


... and the styling on the preview / project renderer is pretty different yet... not sure if that will be addressed separately later, as we need do adjustments there for basically any other block as well.

image

ice9js (commented):
Looks good to me as far as code and functionality is concerned 👍  Thanks!

As for the things @digitalwaveride - the mandatory toolbar toggle is actually missing from all our question blocks at the moment. I suggest we tackle that one separately for all of them and maybe add some sort of shared component for that.

digitalwaveride:
> I suggest we tackle that one separately for all of them and maybe add some sort of shared component for that.

Sounds good to me! 👍



--------------------------------------------
***@***.***
crowdsignal.com | An Automattic Product

digitalwaveride (approved):
João nice!!!
Looking good from my side for now!

As mentioned in our call, some more things to add in the future, but ready to merge for now from my side.

ice9js (approved):
Looks good to me. 🚢  it 😄

PR #93, opened 2021-11-24
Implement the project results view

ice9js:
This patch updates the project results view to feature an iframe which will be used to display current Crowdsignal survey results pages until we reimplement them as part of the SPA.

Solves c/Ks05yCWT-tr.

# Testing

Go to `/project/:projectId/results` to see the iframe.

I set up a proxy on the webpack dev server which will allow the iframe to load locally, but it will currently always be in an unauthenticated state due to missing cookies.

See d70582-code for testing instructions on the platform.

### DIFF
--- a/apps/dashboard/src/components/iframe/index.js
+++ b/apps/dashboard/src/components/iframe/index.js
@@ -1,0 +1,30 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useRef, useState } from '@wordpress/element';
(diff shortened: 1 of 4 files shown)


CGastrell (changes requested):
As mentioned on the diff, the "Editor" tab link still holds the old URL path. I think that's all I have to say about this :)

CGastrell (approved):
:shipit:

PR #96, opened 2021-11-29
Implement a HeaderMeta component

ice9js:
We currently don't have a way to control the page header (`<head>`) from within our application which is slightly less than ideal.  

To help with this, I've implemented a `HeaderMeta` component that is powered by `react-helmet` under the hood. For now it just accepts a single `title` prop and updates the page title accordingly.

It's worth noting that the way `react-helmet` works, if let's say there are multiple nested views where each might want to affect the title, it'll take the data from the last/deepest one which in theory would be the most specific. So that's nice.

# Testing

Verify the title is set correctly for the editor and results views. It should be `Edit Project | Crowdsignal` and `Project Results | Crowdsignal` respectively.


CGastrell (approved):
LGTM!

PR #97, opened 2021-11-30
Fix block alignment on preview and inside the project renderer

ice9js:
This patch fixes the issues with alignment settings having no effect inside the preview or the project renderer.

Solves c/H3p0CZCg-tr.

# Testing

- Add some blocks with alignment options, for example image blocks.
- Verify the alignment settings selected inside the editor are also reflected in the preview or the project renderer.

### DIFF
--- a/apps/dashboard/src/components/editor/style.scss
+++ b/apps/dashboard/src/components/editor/style.scss
@@ -59,6 +59,11 @@ html.interface-interface-skeleton__html-container {
 		position: relative;
 	}
 
+	.block-editor-block-list__layout.is-root-container {
+		padding-left: 0;
+		padding-right: 0;
+	}
+
 	.components-panel__header.edit-post-sidebar__panel-tabs ul {
 			padding: 0
 	}
--- a/apps/dashboard/src/components/editor/style.scss
+++ b/apps/dashboard/src/components/editor/style.scss
@@ -98,6 +98,14 @@ html.interface-interface-skeleton__html-container {
 			margin: 0;
 		}
 	}
+
+	& :where(.wp-block)[data-align=right] > * {
+		float: right;
+	}
+
+	& :where(.wp-block)[data-align=left] > * {
+		float: left;
+	}
 }
 
 .iso-sidebar {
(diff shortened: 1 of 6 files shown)


digitalwaveride:
Thanks Kuba!!!! 🙏 

Things work for me in the "Preview" view, I was adding an image block and a column block, was setting both to "wide width.

The Project Renderer though was acting up for me: Unfortunately not the image block, nor the column block did show up for me: https://digitalwaveride.survey.fm/0B8791C2A67AF3BC?use-project

Expected?

**Editor View:**
image

**Preview View:**
image


**Project Renderer View:**
image

ice9js:
>The Project Renderer though was acting up for me: Unfortunately not the image block, nor the column block did show up for me: https://digitalwaveride.survey.fm/0B8791C2A67AF3BC?use-project

That build is outdated. You can test it locally when you run `yarn workspace @crowdsignal/project-renderer start` and go to `https:crowdsignal.localhost:9001/projectSlug`.

CGastrell (changes requested):
Mmm... not seeing good results on the editor (the image is aligned to the right):
image

And then, on the preview, seems like it's "floating" and the project looks a little weird:
image

At least the renderer is consistent with the preview :D 
image

digitalwaveride:
I can see on the renderer things similar to Christian. If an image is smaller than the "main column width", then the alignment to the left / right is not really working in the editor.

In the project rendered I noticed that the "Full Width" ( Cover) is not working, it only shows in the renderer the "wide width".

ice9js:
I rebased the branch. Unfortunately, after some investigation... it seems the it's in fact the WordPress' image block that's broken.  
I've been able to reproduce the issue in the WP block editor as well but it's only affecting images added from a URL - the only thing we support right now.

When you resize such images, you'll notice it saves negative values for `width` and `height` causing the image to be a random size. As such, I guess the only thing that makes sense for us is to open up an issue/PR in core and disable `centered`, `left` and `right` alignments until it gets fixed.

ice9js:
Reported the issue in WordPress/gutenberg#37374.  
While playing with it a bit more, I noticed it only happens when you select an alignment before resizing an image. If you resize it first and then align it, things work as expected as far as the image block is concerned, so it's something I still need to address on our end.

ice9js:
I'm merging this following our conversation with @digitalwaveride where we decided we'd address anything that's still 'broken' separately.

PR #98, opened 2021-12-01
Disable nav buttons when project is still unsaved

CGastrell:
This PR adds a class/prop combo to disable nav buttons.

The prop `isDisabled` will trigger the class `is-disabled` which sets `pointer-events: none` to the element.

## Test instructions
Checkout and run `yarn workspace @crowdsignal/dashboard start`. Open http://crowdsignal.localhost:9000/project

* see that both "Editor" and "Results" tab buttons don't work while the project is not saved
* make some changes and let the project save (or hit "Save draft"/"Publish")
* see that now the nav buttons work


digitalwaveride (approved):
Works as advertised above! 🚀

PR #99, opened 2021-12-02
Inject a nonce into requests to api.crowdsignal.com when present

ice9js:
This patch adds some compatibility code to append a nonce to any requests to `api.crowdsignal.com` if one is present on `document.body`.  
This enables the application to be used outside of the local development environment with appropriate security measures in place.

# Testing

To see if the code works, modify `dashboard/public/index.html` with the following body tag:

```
<body data-ajax-nonce="foo">
```

When you refresh the page, you should notice requests going out to `api.crowdsignal.com` now have an `x-ajax-api-token: foo` header attached to them.


CGastrell (approved):
Works as advertised! :shipit:

PR #100, opened 2021-12-02
Implement a theme compatibility layer

ice9js:
This patch adds a theme compatibility layer in the form of an `@crowdsignal/theme-compatibility` package.  
The package would contain the so called _compatibility stylesheets_ for each of our base WordPress themes to optimize them for our environment and our blocks.

Right now the setup is pretty basic, every theme would have a single `.scss` file in the `stylesheets` directory named after itself which is then compiled into the respective `dist/[themeName].min.scss`.  
If we ever feel like we need to, we could technically expand this so each theme gets a whole folder but let's keep it simple for now.

One caveat is, since it's a separate package and build, webpack won't hot-reload or even recompile any changes made there when running the dashboard or project-renderer locally.  
For now, it's necessary to run `yarn workspace @crowdsignal/theme-compatibility build` to recompile the css bundles and refresh the page afterwards.  
I'd say this is good enough right now, but I think we might be getting to a point where it might be handy to implement a single command inside the main `package.json` which would bring up both the dashboard and the project renderer as well as set up a watcher on the theme compatibility package.

Because the themes are a separate package and thus part of this repository, their introduction wouldn't change anything about any of the deployment tactics we discussed, so there's that as well.

Solves c/NbfltGwA-tr.

# Testing

Apply the patch and run `yarn workspace @crowdsignal/theme-compatibility build` before testing.
Ensure the following issues are fixed:

- When a `MultipleChoiceAnswer` is followed by a block other than another `MultipleChoiceAnswer`, the margin between the two should be doubled.
- Submit button should no longer appear as full width in the editor. (And should match the preview)
- Increased margin between `QuestionWrapper`s.

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -94,6 +94,7 @@ const Editor = ( { projectId } ) => {
 		'https://app.crowdsignal.com/themes/leven/style-editor.css'
 	);
 	useStylesheet( '/ui/stable/theme-compatibility/leven.min.css' );
+	useStylesheet( '/ui/stable/theme-compatibility/leven-editor.min.css' );
 
 	const handleChangeContent = useCallback(
 		( content ) => {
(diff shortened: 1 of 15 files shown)


digitalwaveride (approved):
🤞

PR #101, opened 2021-12-02
Fill project settings tab

jcheringer:
## Summary 
This PR will start filling the project's sidebar settings with information we already have on the API.

Ref: c/Qga7WIeF-tr

image

## Test Instructions

* Checkout this branch
* Run `yarn install` and then start the dashboard running `yarn workspace @crowdsignal/dashboard start`
* Click the settings icon on the upper right corner and select `Document` tab
* You should something similar with the image above

### DIFF
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -3,33 +3,19 @@
  */
 // eslint-disable-next-line import/named
 import { DocumentSection } from 'isolated-block-editor';
-import { get } from 'lodash';
 import { __ } from '@wordpress/i18n';
-// eslint-disable-next-line
-import { __experimentalGetSettings, format } from '@wordpress/date';
 import { ExternalLink, PanelBody, PanelRow } from '@wordpress/components';
 
-const DocumentSettings = ( { project } ) => {
-	const DATETIME_FORMAT = 'F j, Y H:i';
-	const formatDate = ( timestamp ) =>
-		format( DATETIME_FORMAT, timestamp * 1000 );
+/**
+ * Internal dependencies
+ */
+import { isPublic, getLastUpdatedDate } from '../../util/project';
+import { timestampToDate } from '../../util/date';
 
-	const isPublic = get( project, [ 'content', 'public' ], false );
-	const visibiliy = isPublic
+const DocumentSettings = ( { project } ) => {
+	const visibiliy = isPublic( project )
 		? __( 'Public', 'dashboard' )
 		: __( 'Private', 'dashboard' );
-
-	const publicTimestamp = get(
-		project,
-		[ 'content', 'public', 'timestamp' ],
-		0
-	);
-	const draftTimestamp = get(
-		project,
-		[ 'content', 'draft', 'timestamp' ],
-		0
-	);
-	const lastUpdated = Math.max( publicTimestamp, draftTimestamp );
 
 	return (
 		<DocumentSection>
@@ -40,11 +26,13 @@ const DocumentSettings = ( { project } ) => {
 				</PanelRow>
 				<PanelRow className="project-created-date">
 					<span>{ __( 'Created', 'dashboard' ) }</span>
-					<span>{ formatDate( project.created ) }</span>
+					<span>{ timestampToDate( project.created ) }</span>
 				</PanelRow>
 				<PanelRow className="project-updated-date">
 					<span>{ __( 'Updated', 'dashboard' ) }</span>
-					<span>{ formatDate( lastUpdated ) }</span>
+					<span>
+						{ timestampToDate( getLastUpdatedDate( project ) ) }
+					</span>
 				</PanelRow>
 			</PanelBody>
 			<PanelBody title={ __( 'Permalink', 'dashboard' ) }>
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -39,8 +39,13 @@ const DocumentSettings = ( { project } ) => {
 				<PanelRow>
 					<span>{ __( 'View Project', 'dashboard' ) }</span>
 				</PanelRow>
-				<ExternalLink href={ project.permalink }>
-					{ project.permalink }
+				<ExternalLink
+					href={ project.permalink }
+					title={ project.permalink }
+				>
+					<span className="components-external-link__text">
+						{ project.permalink }
+					</span>
 				</ExternalLink>
 			</PanelBody>
 		</DocumentSection>
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -3,13 +3,118 @@
  */
 // eslint-disable-next-line import/named
 import { DocumentSection } from 'isolated-block-editor';
+import { get } from 'lodash';
+import { __ } from '@wordpress/i18n';
+// eslint-disable-next-line
+import { __experimentalGetSettings, format } from '@wordpress/date';
+import {
+	CustomSelectControl,
+	ExternalLink,
+	PanelBody,
+	PanelRow,
+	ToggleControl,
+} from '@wordpress/components';
 // import { GlobalStylesUI } from '@wordpress/edit-site/src/components/global-styles';
 
-const DocumentSettings = () => (
-	<DocumentSection>
-		Great things will come here
-		{ /*<GlobalStylesUI />*/ }
-	</DocumentSection>
-);
+const DocumentSettings = ( { project } ) => {
+	const dateSettings = __experimentalGetSettings();
+	const formatDate = ( timestamp ) =>
+		format( dateSettings.formats.datetime, timestamp * 1000 );
+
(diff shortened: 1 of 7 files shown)


jcheringer:
@ice9js and @CGastrell though we might need some more work on this, I'm adding you as a reviewer so that we can start talking a look at the things in the current stage.

apps/dashboard/src/components/editor/document-settings.js
```diff
@@ -3,13 +3,62 @@
...
+					<span>{ formatDate( project.created ) }</span>
+				</PanelRow>
+				<PanelRow>
+					<span>{ __( 'Last Updated', 'dashboard' ) }</span>
+					<span>{ formatDate( lastUpdated ) }</span>
+				</PanelRow>
+			</PanelBody>
+			<PanelBody title={ __( 'Permalink', 'dashboard' ) }>
+				<PanelRow>
+					<span>View Project</span>
```
ice9js, line 53: Missing translation.

ice9js, line 53: Shouldn't this be a link/button?

ice9js (commented):
Looks good - I only found a missing translation and I also think the `GlobalStylesUI` comments can be removed for now. I'm not sure that's where these belong when we implement them anyway.

Other than that, I'd love to see some helper functions for reading some of the project's values, but that's somewhat addressed by #75 which is currently on hold and we'll be dealing with it while we're in beta stage. So I think it's best to just leave that as is and refactor everything together later.

apps/dashboard/src/components/editor/document-settings.js
jcheringer: Considering the image attached to the card and the equivalent Permalink implementation on WP.com, this seems to be plain text.

jcheringer:
@ice9js thanks for reviewing it!
I've addressed the comments and, when testing the latest changes, I noticed some required styles adjustments.
Please take a look at the styles changes and let me know if there are any necessary changes to make it more consistent with the project design.

CGastrell:
Weird, followed the instructions and I'm seeing this:
image

Did I miss something?

jcheringer:
> Weird, followed the instructions and I'm seeing this: image
> 
> Did I miss something?

Well, this is what I see too.
Considering what we have for now, this is the result I got.

What is bugging you? Maybe I am missing something 😅

CGastrell:
> What is bugging you? Maybe I am missing something

LOL! I saw the design and I thought it was your screenshot, sorry

digitalwaveride:
Nice João!!! 👍

Three little things:
- The tab itself should be called "Project" not "Document".
- Let's skip the "Last" and just call it "Updated", if there is no other way to avoid the linebreak.
- It seems like there is a little space in front of the "minutes":
image

Thanks so much! 🙏

jcheringer:
> Nice João!!! 👍
> 
> Three little things:
> 
> * The tab itself should be called "Project" not "Document".
> * Let's skip the "Last" and just call it "Updated", if there is no other way to avoid the linebreak.
> * It seems like there is a little space in front of the "minutes":
> 
> image
> 
> Thanks so much! 🙏

Thanks for taking a look Thomas!

Regarding the tab name, I changed the [settings.js](https://github.com/Automattic/crowdsignal-ui/pull/101/files#diff-b7e6c9db5640ef744abed57d40c1f531122b24a98e5b27ffbe7166a89b70ddea) and I'm pretty sure I have seen this text changed during the development. Not sure why it's not working. Any insights @ice9js ?

About the `Last Updated`, to avoid the line break, other than skipping the "Last", the only alternatives I can see are: use an abbreviated DateTime or increase the sidebar width (I guess it's possible).

> * It seems like there is a little space in front of the "minutes":

For this one, indeed, there is a space between the minutes and seconds. I got inspiration from [PostScheduleLabel](https://github.com/WordPress/gutenberg/blob/9df736c678b2e8ef00c355e6cc9fb8092fe25484/packages/editor/src/components/post-schedule/label.js#L8) and it uses  the `@wordpress/date` package, which has this (don't ask me why) [space between the minutes and seconds](https://github.com/WordPress/gutenberg/blob/trunk/packages/date/src/index.js#L125)

As an alternative, we could create our own `DATETIME_FORMAT` constant and make it the format we want to show.

Please, let me know your thoughts. Thanks again! 😃

digitalwaveride:
Let's skip the "Last" for now, simplest solution.

We should keep the date-format that WP uses and not change it for now.

Regarding the space, I am not seeing an issue on WP.com editor... not sure what they are using... 🤷 

image

jcheringer:
> Let's skip the "Last" for now, simplest solution.
> 
> We should keep the date-format that WP uses and not change it for now.
> 
> Regarding the space, I am not seeing an issue on WP.com editor... not sure what they are using... 🤷
> 
> image

Pushed some changes to adjust the "Last Update" label and the date-time format.
About the space between the hour and minutes, I tried to get some help from @CGastrell and, from our understanding the previous implementation was using the same rules WP.com uses. We were not able to identify why the results on our side are different.
This way, I ended up creating our own date-time formatting string. I'm keeping the same format WP.com uses but, this way, I was able to remove the unnecessary space.

digitalwaveride:
Thanks João for looking into the date format! Really appreciated!

I am not sure, maybe just me, but I am seeing this error here:

image

jcheringer:
> Thanks João for looking into the date format! Really appreciated!
> 
> I am not sure, maybe just me, but I am seeing this error here:
> 
> image

I'm not seeing this error in my environment.
Maybe @ice9js can take a look at his env and let us know? 🙏

apps/dashboard/src/components/editor/document-settings.js
```diff
@@ -3,13 +3,60 @@
...
-	</DocumentSection>
-);
+import { get } from 'lodash';
+import { __ } from '@wordpress/i18n';
+// eslint-disable-next-line
+import { __experimentalGetSettings, format } from '@wordpress/date';
+import { ExternalLink, PanelBody, PanelRow } from '@wordpress/components';
+
+const DocumentSettings = ( { project } ) => {
+	const DATETIME_FORMAT = 'F j, Y H:i';
```
ice9js, line 13: NIt: this constant and `formatDate` could be defined outside of `DocumentSettings` since they don't actually rely on anything from the component in their implementation. Let's just put them above it.

apps/dashboard/src/components/editor/settings.js
```diff
@@ -29,7 +29,7 @@ export const editorSettings = {
 			fixedToolbar: false,
 		},
 		toolbar: {
-			documentInspector: __( 'Document', 'dashboard' ),
+			documentInspector: __( 'Project', 'dashboard' ),
```
ice9js, line 32: Technically, this doesn't seem to be valid configuration anymore and the recent version of Automattic/isolated-block-editor expects a bool.  
Let's leave it for now and raise an issue there.

ice9js (changes requested):
I'm not seeing any errors either and am able to test things.

Regarding the document tab name, it seems that [when sidebars were introduced to the isolated block editor](https://github.com/Automattic/isolated-block-editor/commit/38ef55f33e3a9f9ab89af18654db04b097629760) the `documentInspector` setting was made into a boolean and the title is now [hardcoded to 'Document'](https://github.com/Automattic/isolated-block-editor/blob/trunk/src/components/block-editor/sidebar-heading.js#L17).  
Unfortunately there's nothing we can do about that in the very short term so I'd consider that outside of the scope for this PR from now on.

I think the permalink could be made to look a little bit better if we add these styles to the wrapper `<a>`:

```css
text-overflow: ellipsis;
max-width: 100%;
box-sizing: border-box;
overflow: hidden;
display: inline-block;
```

Before:
Screen Shot 2021-12-10 at 2 32 05 PM

After:
Screen Shot 2021-12-10 at 2 29 27 PM

---

I left another small comment as well. Let's get these changes in and merge it.

jcheringer:
> I'm not seeing any errors either and am able to test things.
> 
> Regarding the document tab name, it seems that [when sidebars were introduced to the isolated block editor](https://github.com/Automattic/isolated-block-editor/commit/38ef55f33e3a9f9ab89af18654db04b097629760) the `documentInspector` setting was made into a boolean and the title is now [hardcoded to 'Document'](https://github.com/Automattic/isolated-block-editor/blob/trunk/src/components/block-editor/sidebar-heading.js#L17). Unfortunately there's nothing we can do about that in the very short term so I'd consider that outside of the scope for this PR from now on.
> 
> I think the permalink could be made to look a little bit better if we add these styles to the wrapper `<a>`:
> 
> ```css
> text-overflow: ellipsis;
> max-width: 100%;
> box-sizing: border-box;
> overflow: hidden;
> display: inline-block;
> ```
> 
> Before: Screen Shot 2021-12-10 at 2 32 05 PM
> 
> After: Screen Shot 2021-12-10 at 2 29 27 PM
> 
> I left another small comment as well. Let's get these changes in and merge it.

Thanks for reviewing it again @ice9js!
I've addressed the comments and also used the opportunity to try cleaning the `document-settings.js` up a bit.

ice9js (approved):
LGTM! Thanks!

PR #102, opened 2021-12-06
Fix Image block is not displaying the image on Project Renderer

jcheringer:
## Summary

The purpose of this PR is to fix an issue regarding the images from Image Block not being rendered on Project Renderer view.

Ref: c/szZ0mf3V-tr

## Test Instructions
 
* Checkout this branch
* Run `yarn install` and start the dashboard using `yarn workspace @crowdsignal/dashboard start`
* Create or open a project, add an Image Block and set the image properly.
* Start the Project Renderer using `yarn workspace @crowdsignal/project-renderer start`
* Open the project page inside the Renderer (usually: https://crowdsignal.localhost:9001/[projectId])
* You should see the Image Block rendering the image correctly.

### DIFF
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -109,7 +109,7 @@ const appendInnerBlocks = ( blockName, elements, innerBlocks ) => {
  */
 const createElementsRecursive = ( elements ) => {
 	if ( isEmpty( elements ) ) {
-		return [];
+		return null;
 	}
 
 	return map( elements, ( element ) => {
@@ -117,18 +117,13 @@ const createElementsRecursive = ( elements ) => {
 			return element;
 		}
 
-		const children = [
-			...( element.innerBlocks || [] ),
-			...createElementsRecursive( element.children ),
-		];
-
 		return createElement(
 			element.name,
 			element.props,
 
 			// In practice, if an element is the host for inner blocks
 			// it's children should alwyas be empty here.
-			( children.length && children ) || null
+			element.innerBlocks || createElementsRecursive( element.children )
 		);
 	} );
 };
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -87,15 +87,6 @@ const appendInnerBlocks = ( blockName, elements, innerBlocks ) => {
 	}
 
 	switch ( blockName ) {
-		// case 'core/media-text':
-		// 	const container = find(
-		// 		elements[0].children,
-
-		// 		( element ) => element.props.className.indexOf( 'wp-block-media-text__content' ) >= 0
-		// 	);
-
-		// 	container.innerBlocks = innerBlocks;
-		// 	return;
 		case 'core/cover':
 			const container = find(
 				elements[ 0 ].children,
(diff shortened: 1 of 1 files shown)


packages/blocks/src/render-blocks.js
```diff
@@ -109,21 +109,26 @@ const appendInnerBlocks = ( blockName, elements, innerBlocks ) => {
...
+			...createElementsRecursive( element.children ),
+		];
+
 		return createElement(
 			element.name,
 			element.props,
 
 			// In practice, if an element is the host for inner blocks
 			// it's children should alwyas be empty here.
-			element.innerBlocks || createElementsRecursive( element.children )
```
jcheringer, line 126: While debugging the code I found out that, for the block `core/cover` both `element.innerBlocks` and `element.children` are populated with some objects.

This is preventing the `` from being added to the elements list.

Changing this solved the issue for the image but, if the Image Block has some text over it, the created `<p>` element is being added to the elements list before the `` and, when rendered, the text gets hidden behind the image.

Took a look at WP.com behavior and there, the `core/cover` creates a wrapper with the class `wp-block-cover__inner-container` and adds the `<p>` element inside the wrapper.

Still didn't figure out why this is not happening on our side.

ice9js (commented):
So having looked at it.. I think I see what the problem might be and we actually have a solution ready but it hasn't been used yet.

The core problem is, for some blocks, `core/cover` being a good example, generates a lot of markup in `originalContent` which makes it very difficult to guess where any children/innerblocks should go compared to simple blocks which just have one node.

For example, this is how `originalContent` looks for one of my cover blocks:

```html
<div class=\"wp-block-cover is-light\"><span aria-hidden=\"true\" class=\"wp-block-cover__gradient-background has-background-dim\"></span><img class=\"wp-block-cover__image-background\" alt=\"\" src=\"https://oplatkinatort.pl/wp-content/uploads/2015/03/oplatek-oplatki-na-tort-pinkie-pie.jpg\" data-object-fit=\"cover\"/><div class=\"wp-block-cover__inner-container\"></div></div>
```

Compared to something like a paragraph block which has just: `<p></p>`.

This is a problem that needs to be solved for each block individually which is why we added [that switch statement in appendInnerBlocks](https://github.com/Automattic/crowdsignal-ui/blob/main/packages/blocks/src/render-blocks.js#L89).

Whenever we encounter something like this, that's the place where we should add rules for inserting `innerBlocks` in the correct node depending on the blocks.  

>Took a look at WP.com behavior and there, the core/cover creates a wrapper with the class wp-block-cover__inner-container and adds the <p> element inside the wrapper.
>
>Still didn't figure out why this is not happening on our side.

WordPress renders all its blocks on the backend for public views, and thus doesn't have such issues. In our case, we rely on `originalContent` to carry over the markup from the editor to our public views.

jcheringer:
Hey @ice9js, thanks for the explanation!
I did the changes to handle the `core/cover` block inside the switch statement and it seems to work fine. 👏 
Could you take a look again?

packages/blocks/src/render-blocks.js
```diff
@@ -95,7 +96,16 @@ const appendInnerBlocks = ( blockName, elements, innerBlocks ) => {
 
 		// 	container.innerBlocks = innerBlocks;
 		// 	return;
```
ice9js, line 98: Nit: could we remove this commented code while we're at it? It was here only as a reference but now that we have a working example, I think we can just get rid of it.

ice9js (approved):
Seems to work for me, thanks! Let's maybe just removed the old commented-out example before merging.

packages/blocks/src/render-blocks.js
jcheringer: Sure!

PR #103, opened 2021-12-08
Add behavior to remove empty MC answers using backspace/delete key

jcheringer:
## Summary

The purpose of this PR is to improve Multiple Choice Question block usability by adding the possibility to remove an empty answer by hitting the `backspace/delete` key.

During the development, found an issue in the ` useParentAttributes` hook, which was taking the wrong `clientId` if the block was nested inside other blocks.
That's why we have some 'unrelated' changes on `packages/block-editor/src/util/use-parent-attributes.js`

Ref: c/o9P7UFwf-tr

## Test Instructions

* Checkout this branch
* Run `yarn install` and then start the dashboard running `yarn workspace @crowdsignal/dashboard start`
* Add a Multiple Choice Question
* You should be able to remove an empty answer by hitting the `backspace/delete` key

_Note_: Like on WP.com, you shouldn't be able to remove an empty answer using this method if there are two or less 
 answers on the MC block.

### DIFF
--- a/packages/block-editor/src/multiple-choice-answer/edit-button.js
+++ b/packages/block-editor/src/multiple-choice-answer/edit-button.js
@@ -28,6 +28,7 @@ const EditButtonAnswer = ( {
 	onChange,
 	onReplace,
 	onSplit,
+	onDelete,
 } ) => {
 	const width = attributes.width ? `${ attributes.width }%` : null;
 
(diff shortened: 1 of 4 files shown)


digitalwaveride (approved):
Works works works! 🚀 

Awesome João!

CGastrell (approved):
LGTM! :shipit:

ice9js (approved):
Triple 👍

PR #104, opened 2021-12-09
Hide empty MultipleChoiceAnswer blocks when parent question isn't selected

ice9js:
This patch will hide any empty multiple choice answers when the parent multiple choice question isn't selected.

There are a couple of ways to do this. I opted for one that's slightly more complicated but should give us less headaches down the line.  
The main culprit was that in the editor, each block is wrapped inside another `div.block-editor-block-list__block` which has some spacing styles on it - so hiding just our block doesn't quite cut it.  
This is solved by adding a filter `BlockListBlock` filter to adjust the wrappers' class name based on the block properties. I tried to write it in a way that could be extended to other blocks that might need similar functionality in the future (looking at you poll block 👀 ).

Solves c/4bf109b4-tr.

# Testing

As this introduces a new package, make sure you run `yarn install` before anything else.

- Add a multiple choice question block and leave at least one answer block without a label.
- Any answer blocks without a label should be hidden whenever the wrapping multiple choice question block is not in focus.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -20,6 +20,7 @@
   "main": "src/index.js",
   "dependencies": {
     "@crowdsignal/block-editor": "^0.1.0",
+    "@crowdsignal/block-editor-filters": "^0.1.0",
     "@crowdsignal/blocks": "^0.1.0",
     "@crowdsignal/components": "^0.1.0",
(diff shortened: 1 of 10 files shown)


jcheringer (approved):
Well done! LGTM 🚢

digitalwaveride:
Nice!!!

PR #105, opened 2021-12-09
Include Text Input data when submitting the form

jcheringer:
## Summary

The purpose of this PR is to include the Text Input Form block data when submitting the page by hitting the Submit Button.

## Test Instructions

* Checkout this branch
* Run `yarn install` and then start the dashboard by `running yarn workspace @crowdsignal/dashboard start`
* Include some `Text Input Form` block (and a Submit Button, if not present) on the page and set the necessary properties accordingly
* Hit the `Update` button on the top right (if available) to make sure the changes are saved.
* Start the `project-renderer` by running `yarn workspace @crowdsignal/project-renderer start`
* Open your project (usually https://crowdsignal.localhost:9001/[ProjectId])
* Open the network tab of your Developer Tools
* Fill the input fields and click the Submit Button
* On the Network tab, you should be able to see a request to https://api.crowdsignal.com/v4/projects/[ProjectId]/form and the input fields data should be on the request body.

### DIFF
--- a/packages/form/src/hooks/use-field.js
+++ b/packages/form/src/hooks/use-field.js
@@ -42,7 +42,10 @@ export const useField = ( { name: fieldName, type, value } ) => {
 	const inputProps = {
 		name: fieldName,
 		onChange,
-		value: type === 'checkbox' || type === 'radio' ? value : currentValue,
+		value:
+			type === 'checkbox' || type === 'radio'
+				? value
+				: currentValue || '',
 	};
 
 	if ( type === 'checkbox' ) {
--- a/packages/block-editor/src/text-input/attributes.js
+++ b/packages/block-editor/src/text-input/attributes.js
@@ -1,4 +1,8 @@
 export default {
+	clientId: {
+		type: 'string',
+		default: null,
+	},
 	label: {
 		type: 'string',
 		default: '',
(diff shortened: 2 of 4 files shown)


packages/form/src/hooks/use-field.js
```diff
@@ -42,7 +42,10 @@ export const useField = ( { name: fieldName, type, value } ) => {
 	const inputProps = {
 		name: fieldName,
 		onChange,
-		value: type === 'checkbox' || type === 'radio' ? value : currentValue,
+		value:
+			type === 'checkbox' || type === 'radio'
+				? value
+				: currentValue || '',
```
jcheringer, line 48: Added this to avoid React's uncontrolled to controlled error on the input fields. Is there a better way to handle this?

ice9js: I guess this is about `currentValue` initially being `null` or `undefined` until you start editing it?
Like yourself, I'm not sure if we might be missing something here but that would only become obvious once we start adding more advanced forms. So for the time being, I think this is fine.

ice9js (approved):
Looks good, thanks @jcheringer! 👍

packages/form/src/hooks/use-field.js
jcheringer: > I guess this is about currentValue initially being null or undefined until you start editing it?
Yes, this is the case. Let's keep it this way for now then.

PR #107, opened 2021-12-10
Move project previews to the project renderer

ice9js:
This patch moves the preview responsibility into the project renderer to improve consistency and reduce complexity. The goal of these changes is to eventually completely remove `/project/:projectId/preview` but that in itself is not part of this PR.

Solves c/JApCaglu-tr.

# Testing

This PR depends on D71331-code.

Also make sure to run both `@crowdsignal/dashboard` and `@crowdsignal/project-renderer` when testing this.

- Verify that the projects `Preview` link correctly redirects you to the project renderer.
- Inside project renderer, you should be able to see the latest draft version whenever the url has `?preview=true` appended. Public otherwise.
- The 'public' URL should still work as expected.

Note that there are still a few edge cases which will be addressed separately: for example attempting to access `?preview=true` while not logged in or as a different user will just give you a blank page.  
We need to implement error handling for project-renderer in general and this would be done separately.

### DIFF
--- a/apps/dashboard/src/components/editor/toolbar.js
+++ b/apps/dashboard/src/components/editor/toolbar.js
@@ -69,7 +69,7 @@ const Toolbar = ( { projectId } ) => {
 			? `https://crowdsignal.localhost:9001/${ projectHash(
 					project
 			  ) }?preview=true`
-			: `${ project.permalink }?preview=true`;
+			: `${ project?.permalink }?preview=true`;
 
 	return (
 		<ToolbarSlot className="block-editor__crowdsignal-toolbar">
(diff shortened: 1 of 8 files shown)


jcheringer:
Hey Kuba!
I'm getting an error when trying to access a new Project page (/project)

image

We might need to add some validation on the [toolbar.js](https://github.com/Automattic/crowdsignal-ui/pull/107/files#diff-e102f71cd2c92394edf203566a94f5bbd8e6b05383622cf3d3fa8782c313b07aR67)

ice9js:
Good catch, let me fix that :)

jcheringer (approved):
Looks good, thanks Kuba!

jcheringer (approved):
Looks good, thanks Kuba!

PR #108, opened 2021-12-13
Update Form Fields styling

jcheringer:
## Summary

The purpose of this PR is to make the Text Question block and the Text Input block look a bit nicer on the Project Renderer.

Ref: c/pB6U7n0H-tr

## Test Instructions 

* Checkout this branch, setup and run the `dashboard` and the `project-renderer`
* Open or create a new project and add some Text Question and Text Input blocks
* Click the preview button to open the Project Renderer preview
* The blocks should have some spacing and padding to make them look good

### DIFF
--- a/apps/project-renderer/src/components/app/style.scss
+++ b/apps/project-renderer/src/components/app/style.scss
@@ -1,0 +1,13 @@
+@import '~@wordpress/block-library/build-style/style.css';
+
+.wp-block-columns {
+	margin-bottom: 24px;
+}
+
+.wp-block-columns:not(.is-not-stacked-on-mobile) > .wp-block-column:not(:first-child) {
+	margin-left: 18px;
+}
+
+.crowdsignal-forms-text-input-block {
+	margin-bottom: 24px;
+}
--- a/apps/project-renderer/src/components/app/style.scss
+++ b/apps/project-renderer/src/components/app/style.scss
@@ -5,7 +5,7 @@
 }
 
 .wp-block-columns:not(.is-not-stacked-on-mobile) > .wp-block-column:not(:first-child) {
-	margin-left: 18px;
+	margin-left: 16px;
 }
 
 .crowdsignal-forms-text-input-block {
--- a/packages/blocks/src/text-question/index.js
+++ b/packages/blocks/src/text-question/index.js
@@ -1,7 +1,3 @@
-/**
- * External dependencies
- */
-import { RichText } from '@wordpress/block-editor';
 import classnames from 'classnames';
 
 /**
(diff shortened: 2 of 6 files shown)


packages/blocks/src/text-question/index.js
```diff
@@ -21,10 +17,7 @@ const TextQuestion = ( { attributes, className } ) => {
 
 	return (
 		<QuestionWrapper attributes={ attributes } className={ classes }>
-			<RichText.Content
-				tagName={ QuestionHeader }
-				value={ attributes.question }
-			/>
+			<QuestionHeader>{ attributes.question }</QuestionHeader>
```
jcheringer, line 20: Do we need to use `RichText.Content`?
Using it seems to add a wrapper `<div />` which is affecting some styles.

ice9js: Unfortunately, yes. It doesn't matter with plain text but once you start styling it (bold, italic, etc.) or if you're using emojis it won't render them correctly otherwise :(.  
We actually were using plain `<div>`s on Crowdsignal forms at first and that's how we found out.

Since `QuestionHeader` is a styled component... I wonder if maybe turning it around would help?

```
<QuestionHeader as={ RichText.Content } value={ attributes.question } />
```

ice9js: Actually, just ran into that in #110, without RichText content HTML entities aren't getting decoded correctly.

apps/project-renderer/src/components/app/style.scss
```diff
@@ -0,0 +1,13 @@
+@import '~@wordpress/block-library/build-style/style.css';
```
ice9js, line 1: Nice! Yeah, this looks like it should solve our problems 👍  🙇 

That said, maybe worth leaving a note regarding the future: eventually, we would like to be able to serve Gutenberg separately from our app - similar to how WP does it but it's currently harder for us to do without a WP backend.  
But when we get there, I assume we'd just host a `@wordpress/block-library` `*.css` target somewhere and pull that as a regular stylesheet. Until then, this should serve us well.

ice9js (commented):
Nice, it seems to work for me!

Just one note: this will also require an update to the backend to make sure main.css is included when the page is served from survey.fm.

packages/blocks/src/text-question/index.js
jcheringer: Got it! Thanks for clarifying. I'm reverting those changes.

apps/project-renderer/src/components/app/style.scss
jcheringer: Great! Should we add this note somewhere else?

ice9js: Hmm... I added a reference to this conversation in c/3mQAfwZg-tr.

ice9js (approved):
> I'll need some guidance to add those backend changes. 😬

Sure! This can be merged in the mean time though.

PR #112, opened 2021-12-14
Fix style inconsistencies between the editor and project renderer

ice9js:
This patch fixes any remaining style inconsistencies not yet addressed in #97, #100 or #108. This includes:

- More consistent spacing*
- Updated ...Question blocks to feature a solid background and a shadow by default.
- Added padding at the top of the editor so the toolbar for the uppermost block doesn't get cut-off.

_Note: The spacing remains a little inconsistent between the editor and project-editor. Specifically, in case of nested blocks, some of them might have extra margin in the editor but appear closer together on the public view. I did test the same theme and blocks combination on a WordPress site and got the exact same behavior. See the example below:_

Editor:

Screen Shot 2021-12-14 at 5 10 01 PM

Preview:

Screen Shot 2021-12-14 at 5 10 19 PM

# Testing

Make sure to run `yarn workspace @crowdsignal/theme-compatibility build` before testing.

- Create a new project and proceed to add various blocks.
- There shouldn't be any differences between the editor and the preview/project renderer other than the spacing one mentioned above.

### DIFF
--- a/apps/dashboard/src/components/editor/style.scss
+++ b/apps/dashboard/src/components/editor/style.scss
@@ -62,6 +62,7 @@ html.interface-interface-skeleton__html-container {
 	.block-editor-block-list__layout.is-root-container {
 		padding-left: 0;
 		padding-right: 0;
+		padding-top: 40px;
 	}
 
 	.components-panel__header.edit-post-sidebar__panel-tabs ul {
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -129,8 +129,14 @@ const App = ( {
 	}
 
 	return (
-		<Form name={ `f-${ projectCode }` } onSubmit={ handleSubmit }>
-			<ContentWrapper>{ renderContent() }</ContentWrapper>
+		<Form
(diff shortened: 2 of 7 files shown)


jcheringer (approved):
Nice! Looks good!

digitalwaveride:
Nice @ice9js !!! So fun to play with this!

Besides the image block not addressed in that PR two little things:

#### Font-Size Difference

There is a tiny difference. For example h3 is the same "rem" value, but the total "font-size" varies between editor and renderer. I guess the "base level" is different in the two css files?

image

Difference of paragraph text is even bigger 16px vs 18px

#### Default Styling Question Wrapper

Adjusting the styling of the card a bit... as mentioned in https://trello.com/c/riNvjXjS/1563-%F0%9F%86%95%F0%9F%A7%B1-spacing-inside-of-the-multiple-choice-block-is-off
Shadow off, border-line 1px, color #eee, radius 5px.


Thanks so much Kuba! 🚀

ice9js:
Thanks for the review! Should be ready for another go now @digitalwaveride.

digitalwaveride (approved):
😻 😻 😻 
Thanks Kuba!!!

PR #113, opened 2021-12-15
Enable nonce usage in project-renderer for preview mode to work on survey.fm

ice9js:
This patch updates the `project-renderer` to accept a nonce, similar to the `dashboard` app, so we can make authenticated requests from survey.fm domains.  

Solves c/T9MG7flZ-tr.
Depends on D71582-code.

# Testing

To see if the code works, modify `project-renderer/public/index.html` with the following body tag:

```
<body data-ajax-nonce="foo">
```

When you refresh the page, you should notice requests going out to `api.crowdsignal.com` now have an `x-ajax-api-token: foo` header attached to them.


jcheringer (approved):
Works for me! 👏

PR #115, opened 2021-12-15
Fix Project Title editing

jcheringer:
## Summary

The purpose of this PR is to fix an issue that sometimes overwrites the project title due to concurrent saving requests.

Ref: c/WD04Yw2z-tr

## Test Instructions

* Checkout this branch, setup and run the `dashboard`
* Go to the new project creation page
* You should not be able to reproduce the issue described here: c/WD04Yw2z-tr

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -29,14 +29,17 @@ import { hasUnpublishedChanges } from '../../util/project';
 import './style.scss';
 
 const Editor = ( { projectId } ) => {
-	const [ project, isSaved ] = useSelect( ( select ) => {
+	const [ project, isSaved, isEditDisabled ] = useSelect( ( select ) => {
 		return [
 			select( STORE_NAME ).getProject( projectId ),
 			select( STORE_NAME ).isProjectSaved(),
+			select( STORE_NAME ).isEditDisabled(),
 		];
 	} );
 
 	const projectContent = get( project, [ 'content' ], {} );
+
+	const saveDebounceTimeout = projectId ? 5000 : 500;
 
 	// TODO: need to compare draft/public, offer "restore" drafted content
 	const draftedBlocks = get( projectContent, [ 'draft', 'pages', 0 ], [] );
@@ -55,11 +58,9 @@ const Editor = ( { projectId } ) => {
 		debounce( ( content ) => {
 			try {
 				const blocks = parse( content );
-				const currentProject = project || {};
+
 				saveAndUpdateProject( projectId, {
-					...currentProject,
 					content: {
-						...currentProject.content,
 						draft: {
 							pages: [ [ ...blocks ] ],
 						},
@@ -73,7 +74,7 @@ const Editor = ( { projectId } ) => {
 				// eslint-disable-next-line
 				console.error( error );
 			}
-		}, 5000 ),
+		}, saveDebounceTimeout ),
 		[ projectId, project ]
 	);
 
@@ -98,12 +99,16 @@ const Editor = ( { projectId } ) => {
 
 	const handleChangeContent = useCallback(
 		( content ) => {
+			if ( isEditDisabled ) {
+				return;
+			}
+
 			if ( isSaved ) {
 				changeProjectContent( project );
 			}
 			debounceSave( content );
 		},
-		[ debounceSave, isSaved ]
+		[ debounceSave, isSaved, isEditDisabled ]
 	);
 
 	if ( projectId && null === project ) {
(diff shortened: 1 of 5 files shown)


apps/dashboard/src/components/editor/index.js
```diff
@@ -29,15 +29,18 @@ import { hasUnpublishedChanges } from '../../util/project';
...
 		return [
 			select( STORE_NAME ).getProject( projectId ),
 			select( STORE_NAME ).isProjectSaved(),
+			select( STORE_NAME ).isEditDisabled(),
 		];
 	} );
 
 	const projectContent = get( project, [ 'content' ], {} );
 
+	const saveDebounceTimeout = projectId ? 5000 : 500;
```
jcheringer, line 42: This is a suggestion: When the project isn't created yet,  could you reduce the debounce timeout to save it as soon as possible?
One of the problems I noticed was happening because of the delayed execution of the `debounce`. It was triggering a project update with some outdated data.

apps/dashboard/src/components/editor/index.js
```diff
@@ -55,11 +58,9 @@ const Editor = ( { projectId } ) => {
 		debounce( ( content ) => {
 			try {
 				const blocks = parse( content );
-				const currentProject = project || {};
+
 				saveAndUpdateProject( projectId, {
-					...currentProject,
```
jcheringer, line 60: The `saveAndUpdateProject` method will create or update a project based on its existence. Since it will patch the data, do we need to send data other than what is actually being modified?
This will also help to prevent sending outdated data.

apps/dashboard/src/data/projects/selectors.js
```diff
@@ -21,3 +21,6 @@ export const isProjectPublic = ( state ) =>
 
 export const isProjectSaved = ( state ) =>
 	get( state, [ 'projects', 'isSaved' ], false );
+
+export const isEditDisabled = ( state ) =>
```
jcheringer, line 25: Disabling the editing for a few seconds when we still don't have a saved project helps to avoid parallel project creation requests.

digitalwaveride:
Thanks João!
Things seem to work now 👍 , but I was noticing an irritating "flickering" on the text title, anytime I am clicking the mouse... see video:

https://user-images.githubusercontent.com/5120357/146385450-611ce9af-e48c-43e1-886f-b97dbe0b34e0.mov

digitalwaveride:
Found the issue of the flickering, it was likely related to my sandbox being not properly connected. Flickering is gone now for me!

digitalwaveride (approved):
Works for me! Thanks @João!

apps/dashboard/src/data/projects/selectors.js
ice9js: I'm just curious... couldn't `isProjectSaving` be used to potentially achieve the same result?  
As in, we should send save requests at all if _any_ project is being saved. Regardless if it has an ID or not.

To give more context, you can see there's a bit of inconsistency in project state - some of it requires a project ID while other functions just have a 'global state' like the case here.  
I'd argue the latter is really better considered 'editor' state and would love to clean it up a little once we get the first iteration out of the door.

packages/components/src/editable-page-header/index.js
```diff
@@ -13,17 +13,28 @@ import Form from './form';
  * Style dependencies
  */
 import './style.scss';
+import classnames from 'classnames';
 
-const EditablePageHeader = ( { onChange, text } ) => {
+const EditablePageHeader = ( { onChange, text, isEditDisabled } ) => {
```
ice9js, line 18: Nit: consider renaming the property to just `disabled` - that's inline with all our other inputs/buttons/form components. I'm a fan of keeping things as consistent as possible.

ice9js (approved):
Looks good!
My only question would be about the difference between `isEditDisabled` and `isProjectSaving` - and if we can't just use the latter?

apps/dashboard/src/data/projects/selectors.js
jcheringer: The difference here is that `isEditDisabled` only returns true for new projects, without an Id.
The main reason for this issue was two parallel save requests (without a projectId) being sent to the backend and creating two new projects.
I couldn't think of a way to avoid this behavior without blocking those two requests from happening.
Once we get a projectId, we don't need to care about the parallel requests because the project will be patched every time.

packages/components/src/editable-page-header/index.js
jcheringer: I agree. Thanks for the suggestion. ✅

digitalwaveride (approved):

digitalwaveride:
Let's get it out then?

jcheringer:
> Let's get it out then?

Since @ice9js raised a question about the `isEditDisabled`, I'd like to hear his thoughts about my answer [here](https://github.com/Automattic/crowdsignal-ui/pull/115#discussion_r770932433) before proceeding with the merge.

apps/dashboard/src/data/projects/selectors.js
ice9js: I see. Alright, let's keep it the way it is. Thanks for the explanation.

PR #116, opened 2021-12-16
Update Text Question and Text Input styles

jcheringer:
## Summary

The purpose of this PR is to make the Text Input and Text Question styles more consistent.
We want the paddings to be the same and the Text Question default input height to have room for at least 3 lines of text.

## Test Instructions

* Checkout this branch, setup and run the `dashboard` and the `project-renderer`
* Open or create a new project and add a Text Question and a Text Input block
* The padding for both blocks should be the same and the Text Question input height should have room for at least 3 lines of text

### DIFF
--- a/packages/block-editor/src/text-input/edit.js
+++ b/packages/block-editor/src/text-input/edit.js
@@ -32,7 +32,10 @@ const EditTextInput = ( props ) => {
 	};
 
 	return (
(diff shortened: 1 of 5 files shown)


digitalwaveride:
Thank João!
Things look good for me on the "Preview"... 

image

digitalwaveride (approved):

PR #117, opened 2021-12-16
Include submit button when adding Text Input block

jcheringer:
## Summary

When the user adds the first block that includes "input" or data collecting elements, we need by default add the submit button.
This works already for MC and Question Block but is not happening yet for Text Input block.

## Text Instructions

* Checkout this branch, setup and run the `dashboard`
* Using a project without a submit button, add a Text Input block
* Along with the added block, you should see a Submit Button right after the block


digitalwaveride (approved):
Works for me! 👍 
Thanks João!

PR #118, opened 2021-12-20
Fix project renderer's stylesheet URL when compiling for production

ice9js:
This is a small patch that explicitly lists the host from which to fetch the theme compatibility layer when `@crowdsignal/project-renderer` is compiled for production.  
This is so we can preserve our functionality inside the local development environment but so it won't break when the project renderer is embedded on a different domain than the theme compatibility stylesheet.

Solves c/5AMNs1n1-tr.

# Testing

Theme compatibility layer (`leven.min.scss`) should still load as usual in the local development environment.  
When compiled for production (`NODE_ENV=production`), it should load a stylesheet from `https://app.crowdsignal.com`.


jcheringer (approved):
Works for me!

PR #120, opened 2021-12-21
Add loading state to the Submit Button

jcheringer:
## Summary

The purpose of this PR is to add a loading state to the Submit Button when the user clicks it to send their responses.

Ref: c/vhhutFsG-tr

## Test Instructions

* Open or create a project with some blocks and a Submit Button
* Save it and go to the Preview
* Once you click the submit button you should see it getting into a loading state


packages/form/src/components/form/index.js
```diff
@@ -17,9 +17,13 @@ const Form = ( { children, name, onSubmit, ...props } ) => {
 		[ name ]
 	);
 
+	const { startSubmit } = useDispatch( STORE_NAME );
```
ice9js, line 20: I'm not really a fan of splitting `startSubmit` and `stopSubmit` between different layer of the application and would love to keep them together.  
So either both should be called on the application level (project-renderer) or `onSubmit` could return a promise and we can deal with all of it here.

apps/project-renderer/src/components/app/index.js
```diff
@@ -67,6 +68,9 @@ const App = ( {
 			} );
 	}, [ projectCode, preview ] );
 
+	const { stopSubmit } = useDispatch( STORE_NAME );
+	const formName = `f-${ projectCode }`;
```
ice9js, line 72: If we go with the application level approach (see my other comment below), what do you think about passing `startSubmit` and `stopSubmit` to the submit handler from the `<Form>` component rather than accessing the form store directly here?

So we'd end up with something like `const handleSubmit = ( { data, startSubmit, stopSubmit } ) => ...`

packages/blocks/src/submit-button/index.js
```diff
@@ -1,13 +1,22 @@
...
  * Internal dependencies
  */
-import { Button } from '../components';
+import { Button, ButtonSpinner } from '../components';
+import { useFormState } from '@crowdsignal/form';
 
 const SubmitButton = ( { attributes, className } ) => {
+	const { isSubmitting } = useFormState();
+
+	const RenderedButton = isSubmitting ? ButtonSpinner : Button;
```
ice9js, line 10: I'm not sure I understand... why would we need another `RenderedButton` component instead of just changing the content like:

```
{ isSubmitting ? attributes.label : <Spinner /> }
```

Or basically... I'm not sure there's any reason why the submitting/spinner logic shouldn't be moved inside of this component?

packages/blocks/src/submit-button/index.js
```diff
@@ -1,13 +1,22 @@
...
+
 	return (
-		<Button attributes={ attributes } className={ className } type="submit">
+		<RenderedButton
+			attributes={ attributes }
+			className={ className }
+			type="submit"
```
ice9js, line 16: We should set `disabled={ isSubmitting }` as otherwise it's still possible to re-submit a form that's already being submitted which may or may not cause any weird side effects.

The reason we use `pointer-events: none;` in the editor over `disabled` is it's important to have a good idea what the input will look like when you publish the project - which the `disabled` state would affect.  
But in this case, on the public end, I'd say this no longer applies and if things need be disabled, they should look like they're disabled.

packages/blocks/src/components/button-spinner/style.scss
```diff
@@ -0,0 +1,12 @@
+.crowdsignal-forms-button-spinner {
```
ice9js, line 1: Hmm... I believe we've made the decision to use styled components/`@emotion` for our styles.  
Compatibility stylesheets/editor being the sole exception. Would be nice to keep it consistent.

I guess the benefit would be everything is contained within a single file and we have easy control of where we inject the generated classes. This is not much of an issue now but we will eventually want to move on from our project-renderer/dashboard combo and offer additional apps and embeds where this could impact their practicality.

packages/blocks/src/components/button-spinner/index.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * External dependencies
+ */
+import { Spinner } from '@wordpress/components';
```
ice9js, line 4: In general, we make an explicit effort not to have any dependencies from `@wordpress/components` inside `@crowdsignal/blocks`.   

Most of these require external stylesheets to be loaded for them to work nor do they match our own components or the theme in appearance.   
Even though it might not affect the spinner component itself, I'd rather not make a precedent - and a `Spinner` should be simple enough to reproduce on our end easily.

packages/form/src/hooks/index.js
```diff
@@ -0,0 +1,2 @@
+export * from './use-field';
+export * from './user-form-state';
```
ice9js, line 2: I believe there's a typo in the file name `user-` vs `use-`.

packages/form/src/index.js
```diff
@@ -1,3 +1,5 @@
 export * from './components';
 
-export * from './hooks/use-field';
+export * from './hooks';
+
+export { STORE_NAME } from './data';
```
ice9js, line 5: I'm not sure we should be exporting this - see my comments above on how to deal with that potentially.

ice9js (changes requested):
Thanks Joao! Overall, looks good and like we discussed 👍 . I left a few notes to maybe clean a few things up before we merge it.

packages/blocks/src/submit-button/index.js
jcheringer: To keep the button width while submitting, I'm maintaining the label and just hiding it inside the button. Can you see another way to achieve the same result?

ice9js: Right, but we could just move that logic in here, right? Basically just replace

```
{ attributes.label }
```

With your code from `<RenderedButton>`, so it's something along the lines of:

```
<span className="crowdsignal-forms-submit-button__content">
	{ attributes.label }
</span>
<Spinner />
```

packages/blocks/src/components/button-spinner/index.js
jcheringer: I'm a bit confused here. We have a bunch of cool components inside the `@wordpress/components` but we can't use them to increment our blocks?
I agree that a Spinner is simple but, using it from the WP library would help us make the styles consistent. Wouldn't it?

packages/form/src/hooks/index.js
jcheringer: Nice catch! 👍

apps/project-renderer/src/components/app/index.js
jcheringer: I'm reverting this and handling the submission status inside the Form component.

packages/form/src/components/form/index.js
jcheringer: Yeah, that makes sense. Thanks for pointing it out. 
Going with the approach of returning a promise and handling the status inside the Form component.

packages/blocks/src/components/button-spinner/index.js
ice9js: Right, but the problem is these are best considered 'WP Admin' components. As in, even if you write a WordPress theme or a plugin, you'd only want to use those for the editor/admin UI and not any of the public user-facing views where they would break.

Or think of it as a difference between a button component and a button block - like if we used `<Button>` from `@wordpress/components` instead of our button 'block', things wouldn't look great. So in the same way, we wouldn't want to use `@crowdsignal/components` here either.  
Additionally, for some of these - especially the interactive ones - we might need to provide custom behaviors anyway so they work for our use case.

So while it could be made to work, we opted against it in the hope it would allow us to keep the bundle size smaller as well... which to be honest, I'm not sure if it actually helped much unfortunately.

packages/blocks/src/submit-button/index.js
jcheringer: Got it! Thanks for explaining it.

packages/blocks/src/components/button-spinner/index.js
jcheringer: I see, thanks for clarifying! 👍

jcheringer:
@ice9js thanks for the review. I believe I addressed all the comments. Could you please take a look again? 🙏

packages/blocks/src/components/spinner/index.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import { keyframes } from '@emotion/core';
+import styled from '@emotion/styled';
+
+const spinAnimation = keyframes`
```
ice9js, line 7: Nit: Hmm... let's maybe put that into `packages/styles/src/animations.js` so we have one reference for all our generic animations.

ice9js (approved):
Looks great! Thanks for going through it.

Just left one nit with regard to where to maybe put the animation - feel free to merge afterwards.

PR #122, opened 2021-12-24
Improve project-renderer Responsivity

jcheringer:
## Summary

The purpose of this PR is to update the `project-renderer` style to make it look good on any screen size.

Ref: c/1JLKTgzY-tr

## Test Instructions

* Add a few blocks and play with the screen size
* The layout should look good on all sizes

### DIFF
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -1,13 +1,108 @@
 @import '~@wordpress/block-library/build-style/style.css';
+@import "~@wordpress/base-styles/breakpoints";
+@import "~@wordpress/base-styles/mixins";
+
+:root {
+	--wp--style--block-gap: 1em;
+	--crowdsignal-forms-content-padding: 16px;
+
+	@include break-mobile {
+		--crowdsignal-forms-content-padding: 20px;
+	}
+
+	@include break-small {
+		--wp--style--block-gap: 1.4em;
+	}
+
+	@include break-medium {
+		--wp--style--block-gap: 2em;
+	}
+}
 
 .wp-block-columns {
 	margin-bottom: 24px;
 
+	.wp-block-column:not(:last-child) {
+		margin-bottom: 24px;
+	}
+
 	.crowdsignal-forms-text-input-block {
 		margin-bottom: 0;
+	}
+
+	@include break-small {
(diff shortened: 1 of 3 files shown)


digitalwaveride (approved):
Thanks so much João!

Looking good now!

image

ice9js (approved):
Looks good!

The only concern I have is with breakpoint mixin names: `small` and `medium` are up for interpretation I guess.
I guess a naming scheme like `breakpoint-{ device }[-{ modifier }]` would be slightly more readable. So you'd have stuff like, `breakpoint-tablet` for what's currently presumingly `break-medium`, or you could have `break-mobile-wide` which is slightly wider than whatever we say the baseline for `break-mobile` is. For example it's hard for me to say what `break-small` stands for seeing it next to `break-mobile` and `break-medium`.

Anyway, that's all outside of the scope of this PR, just wanted to mention it. Feel free to merge.

jcheringer:
> Looks good!
> 
> The only concern I have is with breakpoint mixin names: `small` and `medium` are up for interpretation I guess. I guess a naming scheme like `breakpoint-{ device }[-{ modifier }]` would be slightly more readable. So you'd have stuff like, `breakpoint-tablet` for what's currently presumingly `break-medium`, or you could have `break-mobile-wide` which is slightly wider than whatever we say the baseline for `break-mobile` is. For example it's hard for me to say what `break-small` stands for seeing it next to `break-mobile` and `break-medium`.
> 
> Anyway, that's all outside of the scope of this PR, just wanted to mention it. Feel free to merge.

Thanks for sharing your thoughts @ice9js!
I like the idea of making the breakpoints' names more meaningful.
For this implementation, I just reutilized what we have on WP.com but, if it's fine to decouple our breakpoints from theirs, I'll be happy to implement them in a more "Crowdsignal Way".

jcheringer:
Actually, thinking a bit more about the breakpoints' names, I'd say that, giving them a name like `breakpoint-{ device }[-{ modifier }]` would give us the sense of what device this breakpoint targets but we still wouldn't know the specific width the breakpoint handles. We'd end up having to memorize the sizes anyway.

jcheringer:
I'm going to merge it but I'm totally open to discussion and for raising a new PR if we judge necessary.

PR #123, opened 2021-12-28
Refactor saving, publishing and unpublishing project content

ice9js:
This patch addresses a lot of issues that have been bugging us when saving or publishing content, and implements 'unpublish' functionality which was previously missing.

- Our existing 'autosaving' logic based on lodash' `debounce` has been removed in favor of an implementation based on `setTimeout()` which gives us more control. It has been encapsulated within a self-contained `useAutosave` hook to improve the overall readability of itself as well as the main editor component.
- Clicking _'Save Draft'_, _`Publish`_ or _`Update`_ should cancel any autosave delay that might be going on and save the latest changes immediately.
- Editor state has been separated from project state for dashboard data. This again improves readability, but also allows for independent project updates for settings outside of actual content. It also makes it clear there can be many projects, but there's only a single editor instance that's active at a time.
- I updated the **Status & Visibility** part of document settings sidebar to allow for toggling between `public` and `private` (draft) by clicking on the current visibility value and changing it (same flow as WP). Changing to public will also cause the latest changes from the editor to be saved. 
- I updated the toolbar so it accurately reflects the current state and won't cause any duplicated saves where it's possible for older changes to override the current editor state when clicking the save button manually. When a project is public and up to date, _'Draft saved'_ is no longer shown.
- The old _unpublished changes dialog_ has been removed and replaced with an editor notice that a user can either dismiss or use to reload the editor using the latest draft content.
- Fixed an issue where after a project is published, any future changes would cause it to revert to draft.
- I also used this opportunity to split `project.content` into `project.draftContent` and `project.publicContent` in anticipation of #75.
- Fixed an issue where a save request would be made as soon as any content is loaded in the editor. Especially annoying when you have a public project with unpublished changes, where if the public version gets loaded it would immediately overwrite the unpublished draft. Note that [Automattic/isolated-block-editor actually does this on purpose](https://github.com/Automattic/isolated-block-editor/blob/bca504ae1ef98cf1aba136d70e29fc339aa8ec61/src/components/content-saver/index.js#L47).

Screen Shot 2021-12-28 at 5 14 43 PM
Screen Shot 2021-12-28 at 5 13 41 PM

Solves c/UUyCkLXg-tr, c/KswJlRjv-tr, c/nVgKtMVf-tr and c/Q8bjzrVA-tr.

This PR depends on D72304-code.

# Testing

Because the changes are so extensive, please feel free to go beyond the instructions below.
Make sure to test together with D72304-code.

- Try creating a new project, it should save as per usual.
- Make some changes and wait for autosave to kick in after 5 seconds. Your changes should be saved automatically.
- Make some more changes and manually trigger _Save draft_, your changes should be saved immediately and autosave should be cancelled.
- Repeat the above step with _Publish_ instead of _Save draft_. Your changes should be published immediately.
- Verify that any changes made to a project when it's public are 'saved as draft' by default.
- Try reloading the editor page when the project is public but contains unpublished changes. You should see a notice informing you of the fact.
- Clicking _Restore_ on the notice mentioned in the previous step should reload the editor using the latest draft version.
- Open the document sidebar, you should be able to toggle between private (draft) and public state using the visibility setting.
- If you make a change and quickly switch the toggle from private to public, verify that your latest changes were published as well and autosave gets cancelled.
- Opening the editor for an existing project should no longer trigger any save requests until you make changes.
- Preview and public URLs should still work and display the latest draft and public versions respectively.


CGastrell (changes requested):
> Changing to public will also cause the latest changes from the editor to be saved.

This made for an unexpected "Publish": as soon as I changed the visibility to "Public", the change triggered the update. This seems counterintuitive and breaks our "flow". It also disregards the fact that I didn't have any "Submit button".

> I updated the toolbar so it accurately reflects the current state and won't cause any duplicated saves where it's possible for older changes to override the current editor state when clicking the save button manually. When a project is public and up to date, 'Draft saved' is no longer shown.

I got myself in a weird state: Added some paragraph, changed visibility to "Public" (project immediately published). Then made some more changes and suddenly a *draft* save triggered and I ended up with a published project, "Draft saved" label, "Share" button and the "There's a draft version, restore?" notice.

The "Share" button tells me it's *published*, but I still lack a "Submit button". The restore notice showed once, then couldn't make it happen again. But if I reload the editor and do the same (change content on a "Public" project that was not *published*, only changed to *public visibility*) the notice shows.


weird

I left a couple of comments on D72304-code as well

ice9js:
>I got myself in a weird state: Added some paragraph, changed visibility to "Public" (project immediately published). Then made some more changes and suddenly a draft save triggered and I ended up with a published project, "Draft saved" label, "Share" button and the "There's a draft version, restore?" notice.

Turns out a few obscure logic errors remained. To solve it I moved the notice logic into a `UnpublishedChangesNotice` component which and changed the second `useState` argument to `[]`. This ensures the notice logic only gets triggered on the initial render of the editor.  
Previously it could also incorrectly trigger whenever you entered a state where you have a public version and new draft changes and haven't triggered the 'Restore' button yet.  
_Note: because of the added `key` prop, the entire editor is re-initialized when `projectId` or `forceDraft` value changes._

CGastrell (approved):
Works nicely and feels more "stable" :D 

I'm still hesitant regarding the "visibility" as it doesn't respect the "submit button" restriction (it just publishes it with no regards for lawn maintenance) but I think we can tackle that one later

ice9js:
>I'm still hesitant regarding the "visibility" as it doesn't respect the "submit button" restriction (it just publishes it with no regards for lawn maintenance) but I think we can tackle that one later

Noted in c/qj1zYJQO-tr.

PR #124, opened 2021-12-28
Update App styling structure

jcheringer:
## Summary

The purpose of this PR is to refactor the `theme-compatibility` package to create a base stylesheet that will be loaded along with any theme and will be responsible for creating a basic/generic layout for the application, giving it a decent look N feel even without any theme loaded.

## Test Instructions

* Make sure that both, the `dashboard` and the `project-renderer` are running;
* Build the themes packages using `yarn workspace @crowdsignal/theme-compatibility build`;
* The application should work as before and the layout of the `editor` and the `project-renderer` should be quite the same.


apps/dashboard/src/components/app/index.js
```diff
@@ -22,6 +23,10 @@ const allowedRoutes = /^\/(project)\/.*/i;
 const NotFound = () => <div className="app__not-found">404</div>;
 
 const App = () => {
+	useStylesheet( '/ui/stable/theme-compatibility/crowdsignal-editor.css' );
```
jcheringer, line 26: Not sure about adding this one here. 
Tried this aiming to fix c/uprHKmOb-tr.
The Leven theme has some `font-family` definitions and, loading the theme only for the Editor is causing the fonts to be different when switching to the results tab.

apps/project-renderer/src/components/app/index.js
```diff
@@ -119,6 +105,16 @@ const App = ( {
 		);
 	};
 
+	useStylesheet( '/ui/stable/theme-compatibility/crowdsignal.css' );
```
jcheringer, line 108: Left this intentionally commented as a way to see the 'Base Theme'. Need to uncomment this before merging.

apps/dashboard/src/components/app/index.js
ice9js: Right - you're right about the issue, unfortunately that's just how the themes work, also in WordPress land.  
Like you mentioned, since it's only affecting the editor view and our UI is quite limited there outside of the editor itself, I think in this case it'd be ok to just be extra-explicit with styles on `editor/masterbar` and `editor/project-navigation` so they override whatever the theme comes with.

For now let's make sure we've got leven covered, and I guess we might have to update it once in a while once we start introducing new themes.

ice9js (commented):
Hmm... so right off the bat, I think I'd suggest to treat the themes a little differently:

Do you think it'd make sense if we split what currently makes up `crowdsignal/base` into two parts: 

- A shared stylesheet, which would be a minimalistic layer that goes underneath every theme. The idea is this could be `@import`'ed directly inside other themes' stylesheets.
- More specific stuff like colors etc. stays inside `crowdsignal.css` which now also imports that shared SCSS stylesheet.

If we do it that way, that's one less stylesheet we have to load when we boot our app or the editor. And `crowdsignal.css` can become a fully fledged theme, giving us the possibility to adapt or change it without necessarily affecting all other themes (which we still can do using the shared stylesheet).

packages/theme-compatibility/webpack.config.js
```diff
@@ -1,31 +1,31 @@
...
+const process = require( 'process' );
+
+const isDevelopment = process.env.NODE_ENV !== 'production';
 
 module.exports = {
 	mode: 'production',
 	entry: {
-		'leven': './stylesheets/leven.scss',
-		'leven-editor': './stylesheets/leven-editor.scss',
+		'crowdsignal': './src/themes/crowdsignal/base.scss',
```
ice9js, line 11: Nit: since this package is all about themes already, I think we can drop `/src/themes` and have just `/src/crowdsignal`, `/src/leven`, etc.

I guess the shared layer I mentioned would go into `/src/shared` as well then, the only difference being we don't build that one on its own.

jcheringer: > Hmm... so right off the bat, I think I'd suggest to treat the themes a little differently:
> 
> Do you think it'd make sense if we split what currently makes up `crowdsignal/base` into two parts:
> 
> * A shared stylesheet, which would be a minimalistic layer that goes underneath every theme. The idea is this could be `@import`'ed directly inside other themes' stylesheets.
> * More specific stuff like colors etc. stays inside `crowdsignal.css` which now also imports that shared SCSS stylesheet.
> 
> If we do it that way, that's one less stylesheet we have to load when we boot our app or the editor. And `crowdsignal.css` can become a fully fledged theme, giving us the possibility to adapt or change it without necessarily affecting all other themes (which we still can do using the shared stylesheet).

Thanks for the comment @ice9js!
As we discussed, due to some dependencies order, we're not able to import a `shared` file inside the theme stylesheet and load a single file.
I created split the `crowdsignal` theme into two files: a `base` theme, which will be loaded along with the themes we support, and a `crowdsignal` theme itself.

jcheringer:
> Hmm... so right off the bat, I think I'd suggest to treat the themes a little differently:
> 
> Do you think it'd make sense if we split what currently makes up `crowdsignal/base` into two parts:
> 
> * A shared stylesheet, which would be a minimalistic layer that goes underneath every theme. The idea is this could be `@import`'ed directly inside other themes' stylesheets.
> * More specific stuff like colors etc. stays inside `crowdsignal.css` which now also imports that shared SCSS stylesheet.
> 
> If we do it that way, that's one less stylesheet we have to load when we boot our app or the editor. And `crowdsignal.css` can become a fully fledged theme, giving us the possibility to adapt or change it without necessarily affecting all other themes (which we still can do using the shared stylesheet).

Thanks for the comment @ice9js!
As we discussed, due to some dependencies order, we're not able to import a `shared` file inside the theme stylesheet and load a single file.
I created split the `crowdsignal` theme into two files: a `base` theme, which will be loaded along with the themes we support, and a `crowdsignal` theme itself.

CGastrell:
massive!

CGastrell (approved):
This looks good!

My only question would be, how dependent do we become of this? I mean, any change made to the theme will require a re-build from theme-compatibility.

I'd like to see some (at least) minimal doc on the file structure, how they interact with each other and a simple intro/tutorial to create a new custom one (or edit the base one). Not too fancy, but something we can find (and provide) right here on GH

jcheringer:
> This looks good!
> 
> My only question would be, how dependent do we become of this? I mean, any change made to the theme will require a re-build from theme-compatibility.
> 
> I'd like to see some (at least) minimal doc on the file structure, how they interact with each other and a simple intro/tutorial to create a new custom one (or edit the base one). Not too fancy, but something we can find (and provide) right here on GH

Thanks for the review @CGastrell!

What do you mean by "how dependent"? Are you referring to the development environment or production?
Yes, any change on the theme will require a re-build. But, if we're talking about re-building during the development, I think #128 will probably help. If we're talking about production, even before this PR, any change to the theme would require a re-build.

And yes! I agree that some docs would be helpful. Maybe we could address the docs in another ticket? I think it would be nice if could merge this PR soon.

apps/dashboard/src/components/editor/index.js
```diff
@@ -23,13 +22,9 @@ import { editorSettings } from './settings';
...
 import { useAutosave } from './use-autosave';
+import { useStylesheet } from '@crowdsignal/hooks';
 
-/**
- * Style dependencies
- */
-import './style.scss';
-
-const Editor = ( { projectId } ) => {
+const Editor = ( { projectId, theme = 'leven' } ) => {
```
ice9js, line 27: Nit: while it doesn't hurt, we can leave the theme hardcoded for now. Reason being, when we eventually implement theme support on the backend, this will become a property on `project` itself.

apps/project-renderer/src/components/app/index.js
```diff
@@ -20,17 +20,13 @@ import { useStylesheet } from '@crowdsignal/hooks';
...
- */
-import './style.scss';
-
 const App = ( {
 	projectCode,
 	page = 0,
 	preview,
 	respondentId = '',
 	startTime = 0,
+	theme = 'leven',
```
ice9js, line 29: In this case, it might not be a bad idea to leave it as a prop here - on the contrary to the editor - as it would allow us to load things sooner.  
That said, we could actually load the themes synchronously on production and only use this code for the development environment. But that's outside of the scope of this PR.

packages/theme-compatibility/src/base/editor.scss
```diff
@@ -1,9 +1,13 @@
-@import '~@wordpress/components/build-style/style.css';
-@import '~@wordpress/block-editor/build-style/style.css';
-@import '~@wordpress/block-library/build-style/style.css';
-@import '~@wordpress/block-library/build-style/editor.css';
-@import '~@wordpress/block-library/build-style/theme.css';
-@import '~@wordpress/format-library/build-style/style.css';
+@import 'base';
+
+// It seems we don't need those. commented and didn't notice any issues.
```
ice9js, line 3: I think they might all be included in `~isolated-block-editor/build-browser/core.css` anyway. Let's just remove these if they're not necessary.

ice9js (approved):
Looks good to me. Left a few small comments but feel free to merge whenever you're ready :)

apps/dashboard/src/components/editor/index.js
jcheringer: Yeah, the idea here and in the `project-renderer` was to keep the `leven` theme as default, since it's our default theme at the moment, but also provide a way to visualize other themes if you pass the `theme` argument on the URL (`?theme=crowdsignal`).

PR #128, opened 2022-01-04
Add scripts for running and building the entire Crowdsignal UI project

ice9js:
This is a small but effective quality of life improvement which will allow us to run all of our apps as well as build the complete bundle with one command.

- `yarn build`: Builds all relevant bundles and outputs the result inside `/dist` where each package/app goes into its own subfolder names after it.
- `yarn start`: Spins up the development environment. Currently that's `@crowdsignal/dashboard`, `@crowdsignal/project-renderer` and `@crowdsignal/theme-compatibility`.

This should free up a few of those precious shell sessions but on a more serious note, it should also prevent any issues while testing due to forgetting to build/refresh a part of the application.

# Testing

Verify the commands above work as described.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,12 @@
     "webpack-fix-style-only-entries": "^0.6.1"
   },
   "scripts": {
+    "build": "./scripts/build-release.sh",
     "check-licenses": "wp-scripts check-licenses --prod --gpl2",
     "format": "wp-scripts format",
     "lint:js": "wp-scripts lint-js",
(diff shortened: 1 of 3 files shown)


CGastrell (approved):
I couldn't actually test the theme-compatibility, but both start and build ran fine. Editor and renderer work. Nice!

ice9js:
>I couldn't actually test the theme-compatibility, but both start and build ran fine. Editor and renderer work. Nice!

It watches for any changes in that package if you use `npm start`, so they're automatically picked up as well. You still have to refresh manually to see them though.

PR #129, opened 2022-01-05
Add Block Validation

jcheringer:
## Summary

The purpose of this PR is to include a validation mechanism on the Frontend to the blocks that are marked as mandatory.
We need to notify the user about the blocks that are required but are not filled properly.

Ref: c/vzcQbuAs-tr

## Test Instructions

* Open or create a new project;
* Set some blocks as required;
* For Text blocks (Text Input and Text Question) you should see the validation message as soon as you touch the field and leave it in an invalid state;
* You also should see the validation messages for the required blocks when you hit the Submit button without filling required blocks;

### DIFF
--- a/packages/blocks/src/components/form-input-wrapper/index.js
+++ b/packages/blocks/src/components/form-input-wrapper/index.js
@@ -6,7 +6,7 @@ import styled from '@emotion/styled';
 /**
  * Internal dependencies
  */
-import { ErrorMessage } from '../index';
+import { ErrorMessage } from '../';
 
 const FormInputWrapper = styled.div`
 	position: relative;
--- a/packages/blocks/src/components/form-input-wrapper/index.js
+++ b/packages/blocks/src/components/form-input-wrapper/index.js
@@ -6,6 +6,11 @@ const FormInputWrapper = styled.div`
 
 FormInputWrapper.Label = styled.div`
 	margin-bottom: 4px;
+
+	${ FormInputWrapper.className }.is-required & div::after {
+		display: inline;
+		content: ' *';
+	}
 `;
 
 export default FormInputWrapper;
--- a/packages/blocks/src/components/form-input-wrapper/index.js
+++ b/packages/blocks/src/components/form-input-wrapper/index.js
@@ -1,7 +1,23 @@
+/**
+ * External dependencies
+ */
 import styled from '@emotion/styled';
 
+/**
+ * Internal dependencies
+ */
+import { ErrorMessage } from '../index';
+
 const FormInputWrapper = styled.div`
+	position: relative;
 	margin-bottom: 24px;
+
+	${ ErrorMessage.className } {
+		position: absolute;
+		left: 0;
+		bottom: 0;
+		transform: translateY( calc( 100% + 4px ) );
+	}
 `;
 
 FormInputWrapper.Label = styled.div`
--- a/packages/blocks/src/components/form-text-input/index.js
+++ b/packages/blocks/src/components/form-text-input/index.js
@@ -6,7 +6,7 @@ import styled from '@emotion/styled';
 /**
  * Internal dependencies
  */
-import { FormInputWrapper } from '../index';
+import { FormInputWrapper } from '../';
 
 const FormTextInputWapper = styled.div`
 	width: 100%;
--- a/packages/blocks/src/components/form-text-input/index.js
+++ b/packages/blocks/src/components/form-text-input/index.js
@@ -3,6 +3,11 @@
  */
 import styled from '@emotion/styled';
 
+/**
+ * Internal dependencies
+ */
+import { FormInputWrapper } from '../index';
+
 const FormTextInputWapper = styled.div`
 	width: 100%;
 	height: 100%;
@@ -15,6 +20,10 @@ const Input = styled.input`
 	min-height: 40px;
 	padding: 8px;
 	box-sizing: border-box;
+
+	${ FormInputWrapper }.is-error & {
+		border-color: var( --color-error );
+	}
 `;
 
 const FormTextInput = ( { style = {}, ...props } ) => (
--- a/packages/blocks/src/components/question-wrapper/index.js
+++ b/packages/blocks/src/components/question-wrapper/index.js
@@ -8,6 +8,7 @@ import classnames from 'classnames';
  * Internal dependencies
  */
 import { useBorderStyles, useColorStyles } from '@crowdsignal/styles';
+import { ErrorMessage } from '../index';
 
 const StyledQuestionWrapper = styled.div`
 	border-style: solid;
@@ -23,6 +24,15 @@ const StyledQuestionWrapper = styled.div`
 
 	&.has-box-shadow {
 		box-shadow: 2px 2px 8px rgba( 0, 0, 0, 0.23 );
+	}
+
+	&.is-error {
+		border-color: var( --color-error );
+	}
+
+	${ ErrorMessage.className } {
+		margin-top: -16px;
+		margin-bottom: 8px;
 	}
 `;
 
--- a/packages/blocks/src/components/question-wrapper/index.js
+++ b/packages/blocks/src/components/question-wrapper/index.js
@@ -31,8 +31,10 @@ const StyledQuestionWrapper = styled.div`
 	}
 
 	${ ErrorMessage.className } {
-		margin-top: -16px;
-		margin-bottom: 8px;
+		position: absolute;
+		left: 0;
+		bottom: 0;
+		transform: translateY( calc( 100% + 4px ) );
 	}
 `;
 
--- a/packages/blocks/src/multiple-choice-question/index.js
+++ b/packages/blocks/src/multiple-choice-question/index.js
@@ -4,21 +4,35 @@
 import { RichText } from '@wordpress/block-editor';
 import { createContext } from '@wordpress/element';
 import classnames from 'classnames';
+import { isNil } from 'lodash';
 
 /**
  * Internal dependencies
  */
-import { QuestionHeader, QuestionWrapper } from '../components';
+import { ErrorMessage, QuestionHeader, QuestionWrapper } from '../components';
 import { Style } from './constants';
+import { useField } from '@crowdsignal/form';
 
 const Context = createContext();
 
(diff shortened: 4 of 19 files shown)


packages/blocks/src/multiple-choice-question/index.js
```diff
@@ -4,21 +4,35 @@
...
+		name: `q_${ attributes.clientId }[choice]${
+			attributes.maximumChoices !== 1 ? '[]' : ''
+		}`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
+				return 'This question is required';
+			}
+		},
+	} );
+
```
jcheringer, line 29: We may want to refactor this a bit to avoid code repetition on the `MultipleChoiceAnswer`.
Currently, the `useField` hook is used on each answer but, in order to validate it properly, we need to `useField` on the question or pass the `error` status from the answers to the question.

Any thoughts about this one?

digitalwaveride (approved):
Works as advertised!

Thanks João!!!

image

ice9js (commented):
I added a suggestion for how to deal with duplicated/dummy `useField` as well as some other little things to address.

packages/blocks/src/multiple-choice-question/index.js
```diff
@@ -4,21 +4,35 @@
...
 const Context = createContext();
 
 const MultipleChoiceQuestion = ( { attributes, children, className } ) => {
+	const { error } = useField( {
+		name: `q_${ attributes.clientId }[choice]${
+			attributes.maximumChoices !== 1 ? '[]' : ''
+		}`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
+				return 'This question is required';
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/text-input/index.js
```diff
@@ -2,22 +2,38 @@
...
 import { useField } from '@crowdsignal/form';
 
-const TextInput = ( { attributes } ) => {
-	const { inputProps } = useField( {
+const TextInput = ( { attributes, className } ) => {
+	const { inputProps, error } = useField( {
 		name: `q_${ attributes.clientId }[text]`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
+				return 'This field is required';
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/text-question/index.js
```diff
@@ -3,23 +3,35 @@
...
+	QuestionWrapper,
+} from '../components';
 
 const TextQuestion = ( { attributes, className } ) => {
-	const { inputProps } = useField( {
+	const { inputProps, error } = useField( {
 		name: `q_${ attributes.clientId }[text]`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
+				return 'This question is required';
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/text-question/index.js
```diff
@@ -3,23 +3,35 @@
...
+	QuestionHeader,
+	QuestionWrapper,
+} from '../components';
 
 const TextQuestion = ( { attributes, className } ) => {
-	const { inputProps } = useField( {
+	const { inputProps, error } = useField( {
 		name: `q_${ attributes.clientId }[text]`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/text-input/index.js
```diff
@@ -2,22 +2,38 @@
...
+import { ErrorMessage, FormInputWrapper, FormTextInput } from '../components';
 import { useField } from '@crowdsignal/form';
 
-const TextInput = ( { attributes } ) => {
-	const { inputProps } = useField( {
+const TextInput = ( { attributes, className } ) => {
+	const { inputProps, error } = useField( {
 		name: `q_${ attributes.clientId }[text]`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/multiple-choice-question/index.js
```diff
@@ -4,21 +4,35 @@
...
 
 const Context = createContext();
 
 const MultipleChoiceQuestion = ( { attributes, children, className } ) => {
+	const { error } = useField( {
+		name: `q_${ attributes.clientId }[choice]${
+			attributes.maximumChoices !== 1 ? '[]' : ''
+		}`,
+		validation: ( value ) => {
+			if ( attributes.mandatory && ( value === '' || isNil( value ) ) ) {
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/components/form-input-wrapper/index.js
```diff
@@ -1,11 +1,32 @@
+/**
+ * External dependencies
+ */
 import styled from '@emotion/styled';
 
+/**
+ * Internal dependencies
+ */
+import { ErrorMessage } from '../index';
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/components/form-text-input/index.js
```diff
@@ -3,6 +3,11 @@
  */
 import styled from '@emotion/styled';
 
+/**
+ * Internal dependencies
+ */
+import { FormInputWrapper } from '../index';
```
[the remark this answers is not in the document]
jcheringer: ✅

packages/blocks/src/multiple-choice-question/index.js
jcheringer: I've created an `useValidation` hook to be used on the MC Question and also updated the `useField` hook to use it internally.
Please take a look and let me know if the implementation matches what you're suggesting.

packages/blocks/src/multiple-choice-question/index.js
```diff
@@ -4,21 +4,36 @@
...
-import { QuestionHeader, QuestionWrapper } from '../components';
+import { ErrorMessage, QuestionHeader, QuestionWrapper } from '../components';
 import { Style } from './constants';
+import { useValidation } from '@crowdsignal/form';
 
 const Context = createContext();
 
 const MultipleChoiceQuestion = ( { attributes, children, className } ) => {
+	const { error } = useValidation( {
+		fieldName: `q_${ attributes.clientId }[choice]${
```
ice9js, line 21: Hmm, the idea of `useValidation` was it'd just validate the entire form state, so the callback would become something like:

```
useValidation( ( form ) => {
    if ( attributes.mandatory && isEmpty( form[ `q_${ attributes.clientId }[choice]` ] ) ) {
        ...
```

This will also enable us to enforce more complicated rules down the line - like, a question or a form field might only be mandatory when another option is selected.  
So that would make `useValidation` somewhat more universal, while `useField`'s `validation` prop should still only focus on the fields' value.

On another note, I'd be really nice to refactor the form system a little so things like multiple choice fields are actually arrays and not just a comma separated string. One of the reasons being validation - but it doesn't have to be done in this PR necessarily.  
Other reason being, survey submission relies on existing logic but I expect any new endpoints for anything else we do to use proper JSON, in which case this would become somewhat problematic maybe.

packages/blocks/src/components/question-wrapper/index.js
```diff
@@ -24,6 +25,17 @@ const StyledQuestionWrapper = styled.div`
 	&.has-box-shadow {
 		box-shadow: 2px 2px 8px rgba( 0, 0, 0, 0.23 );
 	}
+
+	&.is-error {
```
ice9js, line 29: I believe this has the potential to create some issues:

- If the question's `border-width` is set to `0`, the error border will not show up. 
- While the above is somewhat acceptable, if the question background is then set to transparent, the error messages can look quite confusing.

I think we need to find a way to force the border when an error occurs.

packages/blocks/src/multiple-choice-question/index.js
jcheringer: Some time later (:spongebob)... I finally find some time to go through this.
Could you elaborate a bit more on how you imagined the `useValidation` hook?
From what I understood, it would receive a function as a parameter and, internally, get the form state and call the function passing the form state.
What I didn't understand well is how we'd set the error state in case one of the validations fail. Would we bring this logic to the callback function?

ice9js: Right, so the main difference you're now getting the entire `form` as the argument allowing you to validate any fields from any place within the form.  

As for how you'd set the error state, that callback should return an object with question names as keys and error messages as values.
If a key doesn't exist - that field is valid. Eventually we could maybe even support an array of error messages as values for some more complicated stuff but I'd say that's getting ahead of ourselves a little. For now, if you have multiple callbacks validating the same field, the errors would just override each other. Which is fine, you see one, fix it, then you see the other if it still applies.

CGastrell: I like the idea. Wondering if this *field name* would need to be the same value as in `useField` where the value/name is an encoded name or if we should use a *human readable* or uniquely identifiable name (clientID?)

Given this question, how likely is it that we validate the form values depending on other field values? How would inputs know of each other *by name*?

packages/blocks/src/components/question-wrapper/index.js
jcheringer: Will handle those style changes on c/juGexWOV-tr

packages/blocks/src/multiple-choice-question/index.js
jcheringer: After discussing a few other points with @ice9js, we decide to keep the validation as-is for now and any necessary changes will be handled next.

jcheringer:
Merging this. If we need to do any adjustments, we can do it in a separate PR.

PR #130, opened 2022-01-10
Update Editor and Results header styles

jcheringer:
## Summary

The purpose of this PR is to make the header styles for the Editor and Results tabs consistent.
Currently, when switching between those two tabs, the loaded styles are a bit different, causing a tiny flickering.

Ref: c/uprHKmOb-tr

## Test Instructions

* Open or create a new project;
* Switch between the Editor and the Results tab;
* The header texts should look equal on both tabs and you should not see any flickering;


digitalwaveride (approved):
Smooth! All good, even on Firefox ;) 
Thanks João!

PR #132, opened 2022-01-12
Add type definitions for projects

ice9js:
This patch supersedes the changes from #75 simplifying where possibly and narrowing down to something that's actually valuable.

First of all, I noticed the way our updates work, strict type checking makes little sense hence the introduction of `PartialProject` which constrains an object to contain some but not necessarily all properties of a `Project`.  
For example, `updateProject` expects to be called with an entire `Project` but all other actions work with `PartialProject` data.

Another problem I encountered was conversion from the API response type to our internal `Project` type.  
Fortunately, we're already considering these to be the same thing so no need to overly complicate things, we assume the data returned by the API conforms to `Project`.

One caveat I noticed that was previously bugging us is with the dates, with PHP timestamps being in seconds while JavaScript expects milliseconds. I figured in the long run it would make the most sense to just use ISO strings in API communication in general.  
Hence the type for timestamps has been changed to `string`. ISO strings can still be compared using regular operators and have the added benefit of being compatible/natively parsed by both PHP and JavaScript and pretty much anything else.

Type checking happens by running `yarn check-types` and is optional for now.  It can be enabled by adding `// @ts-check` at the top of the file.

I moved all project util functions into `@crowdsignal/project` as well so we can easily reference them outside of `@crowdsignal/dashbaord` which I believe might come in handy once we start dealing with restrictions/paging/etc. Things that might also be useful in the project-renderer, etc.

_Note: Even when type checking is enabled, it can be deceiving as things like `useSelect` or `useDispatch` don't actually infer types from our selectors and action creators but use `any` by default. Hence everything checks out as long as the code within the file itself isn't contradicting itself.  
It is possible to address this, but it's quite complex and we would need to compile at least some of our code as typescript which I'd like to avoid for the time being. As such, right now it's really more about having a solid reference for these bigger structures._

Depends on D72942-code.

# Testing

Run `npm install` before testing.

- `yarn check-types` shouldn't return any errors.
- Our editor should still work correctly for both new and existing projects.
- Dates in the sidebar should still be formatted correctly.


jcheringer:
@ice9js code looks good to me.
Can we get this rebased so that I can give it a spin? 🙏

ice9js:
Rebased and removed what we decided would be overkill code from `@crowdsignal/dashboard`.
All that's left is a new type-checked `@crowdsignal/project` package which includes the project object description as well as the necessary helper and the related `dashboard/src/util/...` directories have been removed.

In general, I'd say going forward these top-level 'util' files either belong somewhere else more closely related with what they do (editor stuff for example) or a their own separate package (user). Either way, not for now - just noting it here for future reference.

CGastrell (approved):
LGTM! And simple enough :)

ice9js:
>Well, now it's throwing some errors, but none on our packages/files, all coming from node_modules/@emotion
>
>Can we ignore that?

I think that's unrelated and... I guess? There's some conflicts between Gutenberg's `@emotion/*` dependencies and ours and it's something I need to get on top of at some point but it's a battle for another day.

PR #134, opened 2022-01-13
Disable project visibility toggle when project is missing submit buttons

ice9js:
This patch fixes an issue where the visibility toggle in the right sidebar would allow overriding the 'Publish' button when it's disabled because of missing submit buttons on the page.  
With these changes it'll be impossible to publish a page without a submit button.

Solves c/qj1zYJQO-tr.

# Testing

- Create a project and don't add or delete any submit buttons.
- The visibility toggle in the sidebar should be disabled (like the 'Publish' button).
- Add a submit button.
- You should now be able to publish your project through the sidebar toggle.

### DIFF
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -8,7 +8,7 @@ import {
 	PanelBody,
 	PanelRow,
 } from '@wordpress/components';
-import { useDispatch } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
 import { format } from '@wordpress/date';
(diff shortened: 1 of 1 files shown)


jcheringer:
Hey @ice9js, I think we still have a way to get a public project without a Submit button. Here are the steps I did:

* On a public project, with a Submit button;
* Remove the submit button and click the `Visibility` button and change it to Private;
* Note that the popover wasn't closed and let you click on the Public radio button again;

ice9js:
Rebased and fixed the issue with public option still being selectable in some cases.

jcheringer (approved):
Nice! Works for me!

PR #135, opened 2022-01-13
Fix toolbar button styles

ice9js:
A wee fix for toolbar button styles which were getting overridden by base editor styles from the theme compatibility layer.  
As these are already quite specific in what they target and where they used, `!important` shouldn't cause issues in this case and prevent regressions in the future.

Before:

Screen Shot 2022-01-13 at 2 14 57 PM

After:

Screen Shot 2022-01-13 at 2 13 55 PM

# Testing

Verify the toolbar buttons now have the correct appearance.


jcheringer (approved):
Looks good! Thanks @ice9js!

PR #136, opened 2022-01-17
Fix image alignment inside MC Questions

jcheringer:
## Summary

The alignment of the images placed inside an MC Question is inconsistent between the Editor and the Project Renderer.
Also, on the Editor, if the image alignment is set to `left` or `right` the layout gets weird.

Ref: c/gsVmjQs9-tr

## Test Instructions

* Open or create a new project and add an MC Question block;
* Add a few images between the answers and play with the alignment of the images;
* The layout on the editor should look good;
* Also, check the Project Renderer, the layout should be consistent with the Editor layout.

### DIFF
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -49,6 +49,13 @@
 	}
 }
 
+.wp-block-image {
+	.alignleft,.alignright {
+		margin-top: 0;
+		margin-bottom: 0;
+	}
+}
+
 .crowdsignal-forms-form__content {
 	padding: var(--crowdsignal-forms-content-padding);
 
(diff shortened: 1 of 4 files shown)


digitalwaveride:
Nice João!
Issue seems to be fixed! Played with the alignment of the images, and everything worked as expected!!! 🚀 

Just one thing I noticed: The MC block now has pretty big margins between the answer options:

image

Do you see these as well? Or just me?

These are the spacings that are currently live and as they should look like:
image

CGastrell:
I also observe the "extra" space in between the answers.

Image alignment looks good, yet I see the toolbar moving with the alignment, is that expected?
image

digitalwaveride:
> Image alignment looks good, yet I see the toolbar moving with the alignment, is that expected?

Yepp! That is core Gutenberg behaviour:
image

jcheringer:
@digitalwaveride and @CGastrell, I did some adjustments trying to make the spacing more consistent and also moved some generic styles from the Leven theme to the base layer.
Could you take a look again, please?
Thanks! 😃

jcheringer:
@digitalwaveride could you take a look at the inserter issue?
I did some rebasing to get the latest changes from main and now I'm not being able to reproduce the issue anymore.
Could you confirm, please? 🙏

digitalwaveride (approved):
Thanks João!!!

Works for me now! 🚀

PR #137, opened 2022-01-18
Update Isolated Block Editor

ice9js:
This patch updates the Automattic/isolated-block-editor to `2.9.0`, bumps up versions for our tooling and fixes any issues that arose from that - mostly changes in the linter.  

It's worth noting the updated version of the block editor fixes the issue with block previews (examples, block styles etc.) - they should now appear correctly. Albeit still without proper styling.  

# Testing

- Run `yarn install` and `yarn start`
- Everything should build correctly
- Verify our new editor still works, there shouldn't really be any visible difference except for the previews.


digitalwaveride (approved):
Thanks Kuba! 🙏 🙏 🙏 

Previews are not disappearing anymore 👍 !

The "theme styling" is still not in the preview, but I think that was not part of this PR.

image

PR #138, opened 2022-01-19
Fix block previews

ice9js:
This patch fixes remaining issues with block previews: they're now scaled correctly and theme styles are applied.  

As part of this PR I also did a small reshuffling where I extracted are editor UI styles out of `base/editor.scss` and moved them into a dedicated `EditorWrapper` styled component.  
Reason being, the editor itself should never be affected by the theme so rather than hint at that, it's better for the styles to live with the editor itself.

If you're using dev-tools, the previews will be a little laggy as the browser won't cache anything. When viewed with dev-tools disabled, our stylesheets are cached and previews appear quickly and smoothly.

**Note:** Gutenberg will strip any stylesheets loaded from different origins from the previews, so I had to make the theme request go through the webpack proxy as well.

Depends on #137.

Screen Shot 2022-01-19 at 5 22 57 PM

# Testing

Make sure to run `yarn install` unless #137 has already been merged.

- Try hovering over blocks in the inserter panel, previews should appear after a short delay and they should be sized correctly and use the current theme' styles.
- If you select our MultipleChoiceQuestion block, the 'styles' previews in the right sidebar should share the same behavior now.


digitalwaveride (approved):
Works now also on Firefox! 🥳 

image

PR #139, opened 2022-01-19
Fix edit title behavior

CGastrell:
This PR points to a better UX regarding the project title.

Changing the title was triggering a full save which led to some jumpy experience on the editor.
The changeset includes a new state branch/action/selector/reducer:

- Reducer: `state.editor.title` (should it be more specific?)
- Action Type: `EDITOR_TITLE_SET`
- Action: `setEditorTitle( title ) => ( { type: EDITOR_TITLE_SET, title } )`
- Selector: `getEditorTitle()`

The title input change handler simply sets the `title` branch value.

`saveEditorContent` accepts the prop `title` among the options (3rd argument)

`useAutosave` have now a last minute call to `getEditorTitle()` to append to the `saveEditorContent` options argument

Changing the title DOES NOT enqueue an autosave, I could say either I didn't find a good way or a good reason for this so, it's by design.

It also addresses an issue where a newly created project could not be saved except by an autosave (`trying to read id from undefined (project)`)

## Test instructions
Checkout and head to the editor: https://crowdsignal.localhost:8000/project

- Edit the title. Once you *focus out* of the title input you should see "Save draft" became available
- Either add some content to the project (to actually enqueue an autosave) or hit "Save draft": project should save successfully and the title should be set


apps/dashboard/src/components/project-navigation/index.js
```diff
@@ -21,24 +21,29 @@ const Tab = {
...
-		saveAndUpdateProject( projectId, {
-			name: title,
-			title,
-		} );
+	const { setEditorContentChanged } = useDispatch( STORE_NAME );
+
+	const updateTitle = ( title ) => {
+		if ( title !== projectTitle ) {
+			setEditorTitle( title );
+			setEditorContentChanged();
```
ice9js, line 36: Technically we should be able to omit this and update the reducer to also set the flag on `EDITOR_TITLE_SET`.

apps/dashboard/src/components/project-navigation/index.js
```diff
@@ -21,24 +21,29 @@ const Tab = {
...
+
+		return true;
+	};
 
 	return (
 		<div className="project-navigation">
 			<EditablePageHeader
 				onChange={ updateTitle }
-				text={ project?.title || __( 'Untitled Project', 'dashboard' ) }
+				text={ projectTitle || __( 'Untitled Project', 'dashboard' ) }
```
ice9js, line 46: Note that `projectTitle` can be empty if you're loading an existing project into a 'new' editor.  
So this needs to be `projectTitle || project?.title || __( 'Untitled Project', 'dashboard' )`.

apps/dashboard/src/components/project-navigation/index.js
```diff
@@ -21,24 +21,29 @@ const Tab = {
...
+		( select ) => select( STORE_NAME ).getEditorTitle(),
 		[ projectId ]
 	);
 
-	const updateTitle = ( title ) =>
-		saveAndUpdateProject( projectId, {
-			name: title,
-			title,
-		} );
+	const { setEditorContentChanged } = useDispatch( STORE_NAME );
```
ice9js, line 31: Let's concatenate both of these `useDispatch` into a single call - or delete this if you decide to go with my suggestion below where `setEditorContentChanged` would become unnecessary here.

apps/dashboard/src/components/editor/use-autosave.js
```diff
@@ -62,9 +63,11 @@ export const useAutosave = ( projectId, editorView ) => {
 			// Clear any previous autosave timers
 			clearTimeout( autosave.current );
 
+			const title = getProjectTitle();
+
 			// Set a new timer
 			autosave.current = setTimeout( () => {
-				saveEditorContent( projectId, parse( content ) );
+				saveEditorContent( projectId, parse( content ), { title } );
```
ice9js, line 70: Not a huge fan of how this will append the current title value to all update requests. With this affecting stuff like permalinks etc.
Another issue is, editing the title now affects the 'unpublished changes' notice.

On top of that, I believe it's still susceptible to race conditions and the title only gets saved if you edit the content at some point afterwards.

I'm not sure how we can resolve the latter two issues without bumping the save timer logic up into the data layer so it can properly track to-be-saved data coming in from different sources.

apps/dashboard/src/components/editor/toolbar.js
```diff
@@ -24,22 +24,30 @@ import { ToolbarButton } from './styles/button';
...
+		} ).length > 0,
+		select( 'core/block-editor' ).getBlocks(),
+		select( STORE_NAME ).isEditorSaving(),
+		select( STORE_NAME ).isEditorContentSaved(),
+		select( STORE_NAME ).getEditorTitle(),
+	] );
 
 	const saveProject = () =>
-		saveEditorContent( project.id, parse( serialize( editorContent ) ) );
+		saveEditorContent( project.id, parse( serialize( editorContent ) ), {
```
ice9js, line 44: This actually already was a drawback of us not having a reference of the to-be-saved changes - we have to extract blocks manually anytime we want to call `saveEditorContent`.  
Exactly the problem that having the timer in the state layer would solve.

ice9js (changes requested):
I was leaning towards allowing it on the premise it will be refactored next time we need to touch any of these things... but seeing some of the problems with this approach - some stemming from how `useAutosave` wasn't quite an ideal solution itself - I think we just need to get on with it.

apps/dashboard/src/components/project-navigation/index.js
CGastrell: I tried it with an existing project and it worked fine. Not sure what you mean by *'new' editor*

apps/dashboard/src/components/editor/use-autosave.js
CGastrell: > Not a huge fan of how this will append the current title value to all update requests.

Why not?

> With this affecting stuff like permalinks etc.

That is a problem of backend and should be dealt with there. From the editor we just send the usual payload of data to be saved

> editing the title now affects the 'unpublished changes' notice

Yeah, didn't think about that 

> I believe it's still susceptible to race conditions...

Not sure "what" race conditions you're talking about.

> and the title only gets saved if you edit the content at some point afterwards.

I'll repeat this PR's intro: *I could say either I didn't find a good way or a good reason for this so, it's by design.*

> I'm not sure how we can resolve the latter two issues without bumping the save timer logic up into the data layer so it can properly track to-be-saved data coming in from different sources.

Well, if this is definitely the goal, then we can just move the things we want to our *own* state branch and do all this data layer dancing that you see fit (title, editorContent, timers/autosave/notices)

apps/dashboard/src/components/editor/toolbar.js
CGastrell: Oh, so this is like "pre-existing condition" sort of speak.

apps/dashboard/src/components/editor/use-autosave.js
ice9js: It is, and it'll also help with stuff like having to fetch the content manually from the editor anytime you want to force a save.  
Will also be useful for stuff like paging for the same issue that now, if you would like to add a new page, you need to fetch the editor content manually because we're updating the entire thing and we don't want any debounced changes waiting to be saved to get lost when we switch context.

CGastrell:
This was overtaken by #142

PR #140, opened 2022-01-20
Change state editor content

CGastrell:
This PR is a first step towards managing content from our own state branch (as opposed to take whatever the isolated editor throws at us).

The PR has been branched off #139 and it carries some of its not-so-convincing strategies, but retains functionality and allows for collecting all the to-be-saved data on `state.editor` branch.

Autosave functionality has been removed, but it could be tailed after/inside useEditorContent hook (introduced on this PR). I tried to chain autosave but didn't dedicate a lot of time to it, any hint, appreciated.

`useEditorContent` (naming patent pending) is a simple hook to take editor callback and update our state's `editor.content` branch. Saving mechanism now takes content from that branch to send it to the API.

NOTES:
- syncing to our state branch is debounced
- third option on `saveEditorContent` is spread/destructured into the data for the API, meaning any prop there is going to be sent as a payload prop. We might want to prefer doing (as it was) a *per option* assignment, but so far everything we put there was meant to be sent with the payload.


apps/dashboard/src/components/editor/use-autosave.js
```diff
@@ -19,8 +19,9 @@ export const useAutosave = ( projectId, editorView ) => {
 
 	const autosave = useRef();
 
-	const [ isEditorContentSaved ] = useSelect( ( select ) => [
```
ice9js, line 22: If the file isn't used, let's either remove it or at least remove those changes from the diff.

apps/dashboard/src/components/editor/index.js
```diff
@@ -21,8 +21,9 @@ import EditorLoadingPlaceholder from './loading-placeholder';
 import { editorSettings } from './settings';
 import Toolbar from './toolbar';
 import UnpublishedChangesNotice from './unpublished-changes-notice';
-import { useAutosave } from './use-autosave';
+import { useEditorContent } from './use-editor-content';
 import { useStylesheet } from '@crowdsignal/hooks';
+import { EDITOR_VIEW_AUTO, EDITOR_VIEW_DRAFT } from '../../util/editor';
```
ice9js, line 26: `../../util` is literally about to be removed in #132. Regardless of that, these should just be in `src/components/editor/constants.js` or something - don't see how anything else than the editor component would need to use these.

apps/dashboard/src/components/editor/use-editor-content.js
```diff
@@ -0,0 +1,58 @@
...
+	} = useDispatch( STORE_NAME );
+
+	// If projectId or editorView changes, reset 'ready'
+	useEffect( () => {
+		setReady( false );
+	}, [ projectId, editorView ] );
+
+	return useCallback(
+		// using debounce here so we don't pound the state synchronization
+		debounce( ( content ) => {
```
ice9js, line 33: Pound it! The whole idea of moving the synchronization into the state layer is that the state layer is always up to date.  
If anything we can just save `content` to the editor state as is and `parse()` it before we push it to `saveAndUpdateProject`, so we don't have to do that part on every keystroke.

apps/dashboard/src/components/editor/use-editor-content.js
```diff
@@ -0,0 +1,58 @@
...
+				// editorView changes via setForceDraft to EDITOR_VIEW_DRAFT, meaning a restore intent
+				if ( editorView === EDITOR_VIEW_DRAFT ) {
+					restoreEditorContent( content );
+				}
+				setReady( true );
+				return;
+			}
+
+			// Mark editor content as changed
+			setEditorContentChanged();
```
ice9js, line 48: I guess now that we have actions for saving the actual changes, we don't need `setEditorContentChanged()` anymore - the reducer can just update the flag when any of the other actions is dispatched.

apps/dashboard/src/components/editor/use-editor-content.js
```diff
@@ -0,0 +1,58 @@
...
+		// using debounce here so we don't pound the state synchronization
+		debounce( ( content ) => {
+			// Isolated block editor forces a save as soon as the editor content has loaded.
+			// Ignore the first save and set 'ready' to true.
+			//
+			// https://github.com/Automattic/isolated-block-editor/blob/bca504ae1ef98cf1aba136d70e29fc339aa8ec61/src/components/content-saver/index.js#L47
+			if ( ! ready ) {
+				// editorView changes via setForceDraft to EDITOR_VIEW_DRAFT, meaning a restore intent
+				if ( editorView === EDITOR_VIEW_DRAFT ) {
+					restoreEditorContent( content );
```
ice9js, line 41: Why do we need this? As in, you can only restore the draft if you haven't made any changes. Meaning, the editor content state would also be empty at that point. What am I missing?

apps/dashboard/src/components/editor/use-editor-content.js
```diff
@@ -0,0 +1,58 @@
...
+	}, [ projectId, editorView ] );
+
+	return useCallback(
+		// using debounce here so we don't pound the state synchronization
+		debounce( ( content ) => {
+			// Isolated block editor forces a save as soon as the editor content has loaded.
+			// Ignore the first save and set 'ready' to true.
+			//
+			// https://github.com/Automattic/isolated-block-editor/blob/bca504ae1ef98cf1aba136d70e29fc339aa8ec61/src/components/content-saver/index.js#L47
+			if ( ! ready ) {
```
ice9js, line 38: You certainly do not want to debounce this condition!

apps/dashboard/src/data/editor/reducer.js
```diff
@@ -42,7 +49,32 @@ const hasUnsavedChanges = ( state = false, action ) => {
 	return state;
 };
 
+const title = ( state = '', action ) => {
+	if ( action.type === EDITOR_TITLE_SET ) {
+		return action.title;
+	}
+
+	if ( action.type === PROJECT_UPDATE ) {
```
ice9js, line 57: The editor state should only contain changes waiting to be saved. I'd remove this.
it's also confusing because now this gets updated whenever `PROJECT_UPDATE` gets called for any project at any time.

apps/dashboard/src/data/editor/reducer.js
```diff
@@ -8,9 +8,13 @@ import { combineReducers } from '@wordpress/data';
  */
 import {
 	EDITOR_CONTENT_CHANGED_SET,
+	EDITOR_CONTENT_RESTORE,
 	EDITOR_CONTENT_SAVE,
 	EDITOR_CONTENT_SAVE_ERROR,
 	EDITOR_CONTENT_SAVE_SUCCESS,
+	EDITOR_CONTENT_UPDATE,
+	EDITOR_TITLE_SET,
+	PROJECT_UPDATE,
```
ice9js, line 17: One action I'd add is `EDITOR_INIT` - which would 'reset' this tree when `<Editor />` is first loaded or `projectId` changes.  
It might also be useful to have a `state.editor.projectId`.

So you'd initialize the editor with something like `initializeEditor( projectId )`, that way we can later also reference that in what project the changes should be saved to.

ice9js (changes requested):
Nice, this is going in the right direction. I still have some questions but shouldn't take much more.

apps/dashboard/src/components/editor/index.js
CGastrell: Gotcha, didn't know where to put it, will move those.

apps/dashboard/src/components/editor/use-editor-content.js
CGastrell: Well, when we want to restore the content we use the action from the notice to `setForceDraft( true )`, which changes `editorView` parameter from `auto` to `draft` and picks the draft content over the public one. This param is being *watched* by the `useEffect` at line 26 and, as it changes, sets `ready` to false, triggering the re-render/eval of the hook. When we get to this point with the 2 conditions (ready === false && editorView === draft) it means a restore is going on, so we restore the content, set ready to true and abort the callback.
But now I'm guessing you knew this and maybe your question aims to some different explanation?

Anyways, something broke and I'm not getting the desired effect (restore should "rollback" to draft and leave the app in a "saved" state, meaning the "Save draft" button should be disabled. This was happening when I pushed the changes and now it's not, need to figure it out.

This is also valid for your comment/question above:

> I guess now that we have actions for saving the actual changes, we don't need setEditorContentChanged() anymore - the reducer can just update the flag when any of the other actions is dispatched.

I think it all started to fail now that I've removed the `debounce`d call.

apps/dashboard/src/data/editor/reducer.js
CGastrell: What?!

apps/dashboard/src/data/editor/reducer.js
CGastrell: I'm not sure how to replace this one. The goal here is to catch when the project has been loaded from API and set the title from that source... where would you put it (other than in a control)?

I think part of the issue here is that, on the editor state branch, we treat things separately. So, a CONTENT_UPDATE won't/shouldn't affect `editor.title`, yet a PROJECT_UPDATE (only triggered after save/load) should.

ice9js: For now we can just do `projectTitle || project.title || default` in the project navigation thingy for the title value.  
We might want to talk about it some more, agreed... 
Perhaps just add dedicated selectors inside editor state so we don't have to solve these manually wherever they appear? Where it takes the value from editor state or falls back to `state.projects.id` using the projectId currently in `state.editor`... This is getting so confusing so quickly, lol.

apps/dashboard/src/data/editor/reducer.js
ice9js: This isn't actually necessary right now, although having current `projectId` inside `state.editor` would be nice.  

What I'm referring to is a situation down the line, you edit one project, exit, eventually make your way to editing another one, or even the same one - we need to make sure `state.editor` is reset to the defaults/empty before that second editor gets loaded. So I guess a general `initializeEditor( projectId )` would take care of any setup loading a new project might require - but most of it is not a problem for now anyway so maybe let's not overthink this too much.

apps/dashboard/src/components/editor/toolbar.js
```diff
@@ -24,22 +24,30 @@ import { ToolbarButton } from './styles/button';
...
+		} ).length > 0,
+		select( STORE_NAME ).getEditorContent(),
+		select( STORE_NAME ).isEditorSaving(),
+		select( STORE_NAME ).isEditorContentSaved(),
+		select( STORE_NAME ).getEditorTitle(),
+	] );
 
 	const saveProject = () =>
-		saveEditorContent( project.id, parse( serialize( editorContent ) ) );
+		saveEditorContent( project?.id, parse( editorContent ), {
```
ice9js, line 44: So now that editor content is stored on `state.editor` we should be able to just call `saveEditorContent()` here. No need to fetch the data manually.

Same goes for `publishProject()`.

apps/dashboard/src/data/editor/reducer.js
```diff
@@ -42,7 +47,28 @@ const hasUnsavedChanges = ( state = false, action ) => {
...
+		return action.title;
+	}
+
+	return state;
+};
+
+const content = ( state = '', action ) => {
+	if (
+		action.type === EDITOR_CONTENT_UPDATE ||
+		action.type === EDITOR_CONTENT_RESTORE
```
ice9js, line 61: I think `RESTORE` should just set this to `''` rather than `content`, right ? Or not do anything at all.

ice9js (changes requested):
Just a few minor things but I think this is almost good to go.

apps/dashboard/src/data/editor/reducer.js
CGastrell: The restore receives the new content from the project.draftContent (meaning, it restores to whatever you provide), the main difference is that a restore (like an initial load) skips triggering the isolated-editor `onSaveContent` callback.

apps/dashboard/src/components/editor/toolbar.js
CGastrell: If we want `saveEditorContent` to be callable without any params, all the content, title and extra options should come from somewhere. Content and title are now at `state.editor`, but `public` is something we do *on demand*.

I think this assumes that we already have the project ID stored on `state.editor`?

ice9js: Yeah. I think it makes sense to keep `public` as an option rather than in editor state as well - like, publishing in itself is more of an action than it is data.

CGastrell:
This was overtaken by #142

PR #141, opened 2022-01-21
Project editor page navigation

ice9js:
This PR updates the editor with a page navigation column which enables users to create multi-page projects.  
The column features a live preview for each page as well as its number. Drag-and-drop reordering will be added in another PR.

For now, to contain the scope of this PR and while we figure things out on the backend, the list will not include the 'thank you page' which still remains uneditable.

crowdsignal localhost_9000_project_2686135

# Testing

- When you open the editor, it should now feature a _Pages_ column on the left side. Creating a new project should start with a new page.
- You can add new pages by using the `+` button in the page navigation column. The newly added page is immediately loaded into the editor.
- Clicking on the page will reload the editor with the selected page content.
- Hovering over a page in the page navigation column should also display a trash icon to delete it.
- Clicking the delete button should delete the page. If the page was selected, the editor will move to the next page or if the deleted page was the last page, it'll move to the previous page.
- The delete page button is disabled if there's only one page in the project.
- Adding and removing pages is disabled while the editor is saving.
- The publish button is now disabled if any page is missing a submit button.


CGastrell (approved):
Massive work!

You've moved a lot of things into the state/dispatch realm, LGTM. The editor changeset flow is little confusing, but it gets the work done.

Regarding the use of multiple stores... on one hand I'm ok with it.
On the other... it would change the complexity to a different level: we would struggle figuring out which is the right store. Nothing wrong with that, just wondering if that is an actual solution/problem for us.

If it's just for the name collision on the selectors, I'd say let's get creative on those and be done with it.

Let's discuss more use cases for it and then we can come up with a decision.

Now, for a final note on this PR, I'm guessing a lot of backend changes are in order. Saving seems to work fine, but surely the renderer is getting confused or not even know how to submit one page or what to do with the last one. The submit button restriction needs to be tuned up to distinguish _on what pages_ it should act upon (I'm guessing, once you have more than one page, even with no questions on a page we need to force a submit button, the only exception would be the "thank you" page which is not yet available to edit)

CGastrell:
One thing I've noticed is "Update" is always available (even after updating the public version of the project)

ice9js:
I believe I addressed most of the issues - the only one remaining would be the ellipsis menu for pages, but that'll be done in a separate PR.

I also ported the reducer/selector structure from #146 back into dashboard's editor state. As previously discussed with #cgastrell, the changes introduced do make the code much more understandable and less error prone but we don't necessarily want to to extract the editor state to a separate package just yet.

digitalwaveride (approved):
Nice Kuba! 

Thanks so much for addressing the issues, did not see the Ghost button anymore... and previews look great now! 👍 

From my side all good here ... as the preview/render view regarding getting to the next page is not yet working, we should just not yet push it "live" into the closed beta, but merging it ... up to you!

Oh, one nitty-gritty thing that I noticed, just a quick text change for now ... (without additional review):

### Publish Button deactivated notification

This message shows if you have actually already a button on your page, but you have a submit button missing on some other page than the current page you are looking at.

image

A text change like this... could do the trick for now... (... later as discusse we would need some more sophisticated solutions for our users)

`Please make sure that all your pages have at least one submit button in order to publish your project.`

PR #142, opened 2022-01-24
Move autosave logic into the data layer

ice9js:
This PR takes over from #140, implements the missing autosave and cleans up the editor state. Key changes are:

- `state.editor` now contains a copy of all *unsaved changes*.
- I implemented controls for dealing with the autosave timer (`@wordpress/data` doesn't allow middleware). The timer now covers the entire `state.editor` subtree.
- `ProjectNavigation` now expects a handler for saving the title to be provided externally due to different handling on editor and results tabs.
- Due to `state.editor` having a reference for all unsaved changes, it's no longer necessary to query everything manually when you want to... say, publish a project.
- `state.editor.content` is stored as a string and only re-parsed when it's actually about to be saved.
- `saveEditorChangeset()` takes all unsaved changes from `state.editor` and proceeds to call `saveAndUpdateProject` with them, resetting them to defaults/empty afterwards.
- `redirect()` call for new projects has been moved from `saveAndUpdateProject()` to `saveEditorChangeset()` where it makes much more sense. Allowing us to use the former in more than just the editor context - for example, changing the title on the results tab.

This should solve all of our remaining data synchronization issues in the editor view.

# Testing

- Check manual- and auto-saving behavior for new and existing projects.
- Changes to the project title in the editor tab should now also be debounced by the same autosave timer as the editor content.
- Changes to the project title in the results tab will be saved immediately.
- When you have unsaved changes in the editor and go to the results tab, the changes should be saved in the process.
- Hitting the publish button (or the publish option in the visibility toggle) will save any unsaved changes along with making the project public.
- Reverting to _private_ in the visibility toggle will trigger a save immediately, but will not include any changes - it will only remove the public version of the project, with any ongoing autosave remaining unaffected.


CGastrell (approved):
Looks good, editor and saves work nicely! Thanks for solving that autosave riddle.

The only thing that is misbehaving is the "Results" tab. I keep getting the 404 thing:
image

Going to "Results" and back leaves the buttons in a "show all" state
image
After auto save:
image

That said, I think this change is big enough, we should split any undesired effect to a new PR. If you insist in dealing with it here though, I'll review it again

PR #143, opened 2022-01-26
Fix editor/base.scss stylesheet

ice9js:
This is a small patch to clean up styles which already have been moved to [editor/styles/editor.js](https://github.com/Automattic/crowdsignal-ui/blob/main/apps/dashboard/src/components/editor/styles/editor.js), but I forgot to remove from [base/editor.scss](https://github.com/Automattic/crowdsignal-ui/blob/main/packages/theme-compatibility/src/base/editor.scss) in #138. By being duplicated, they're causing conflicts.

# Testing

- The changes should not affect the editor in its current state, functionally or visually.

### DIFF
--- a/packages/theme-compatibility/src/base/editor.scss
+++ b/packages/theme-compatibility/src/base/editor.scss
@@ -29,42 +29,6 @@
 }
 
 .editor .iso-editor {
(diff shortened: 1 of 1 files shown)


CGastrell (approved):
IPF

PR #145, opened 2022-01-28
Add image upload on image blocks

CGastrell:
This PR adds the functionality to upload images to Crowdsignal's media library when inserting an image. It's still a WIP as things might need a little grooming or better implementation.

Core blocks still show a big dependency on the WP environment/data, the isolated-block-editor provides a couple of filters to opt out of some core functionality and replace it with a customized version of our own. That said, the core/image block still attempts to request to WP's public API.

Commits links (for cleaner view):

- https://github.com/Automattic/crowdsignal-ui/commit/476f9e0b12ccbf64ecd3718d3e3086dcb8fb7caa adds a new option `skipGlobalHeaders` on http package. The option should be an array of headers to be omitted from the package's `globalHeaders`.
- https://github.com/Automattic/crowdsignal-ui/commit/fa8374d3ef0839fceadff205028fcb5ff9bfce0b Adds the wrapped call to `/media` API endpoint.
- https://github.com/Automattic/crowdsignal-ui/commit/c61309244af10e9490587916ddc2781f88b9a679 New filter/HOC to add `linkDestination` to core/image blocks. The prop is listed as deprecated, but the core block will check for its value and fallback to `window.wp` things. I left a note on the code, hopefully at some point we might also deprecate/remove it.
- https://github.com/Automattic/crowdsignal-ui/commit/f680679111d41ca949c7256f61eb3e5186d0cdaf Adds the file browser modal handler/callback. Once provided as an editor setting, this handler is called with a *normalized* version of the selected file.
- https://github.com/Automattic/crowdsignal-ui/commit/8be2de27fc0e0bb2f7d09bcdc3e45834d10a0680 Make use of new filters: with-media-link-destination and media-upload. The latter will replace the editor's default MediaUpload component, responsible for the "Open Media Library" button shown right beside the "Upload" button. For now we just use a `null` component.
- https://github.com/Automattic/crowdsignal-ui/commit/33ee9d45eff7dbe794199d056a07ae84ef6dac79 Adds the settings to enable the custom `uploadMedia` handler.

Notes:

- Some settings do not take any effect (`allowMimeType`, `hasUploadPermission`), I'm still figuring a couple of things and extra eyes could come in handy. I think most of the work can be done on the uploadMedia handler, fully customizing it to work with our needs.
- The interaction between MediaUpload and MediaPlaceholder components is still fuzzy to me. In the end, I think we'll need to create our own custom ones for both. 
- `uploadMedia` is used both via internal dependency and through a `select( CORE_STORE ).getSettings().mediaUpload`
- ~~`core/image` still triggers a request to `wp/v2/media` with (so far) no side effects, but needs attention~~ https://github.com/Automattic/crowdsignal-ui/pull/145/commits/16b570c1e21ed994fc2687ef5d462e1136b8c091: Removing the `id` prop from the media object makes the block not try to get it from WP API
- API endpoint ( D73950-code ) is right now returning some hardcoded media payload, ~~but it should be better defined and then mapped on the client side to fit the needs of the core/image block.~~ Diff was updated with a better representation of the media object
- Need to figure out the allowedTypes/mime settings.

Some handy refs:

- MediaPlaceholder https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/media-placeholder
- MediaUpload https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/media-upload
- Some core/image shenanigans https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/edit.js#L224

## Test instructions

**Patch D73950-code on your sandbox, make sure your machine is connecting to sandbox and not prod.**
Create/edit a project. Insert an image block. You should see the "Upload" button. Use it.
Go to you Crowdsignal dashboard and edit/create some poll/survey. Use the "Add media" buttons on a question to open the media library popup, you should see the image uploaded on the project among your media files.
You can use Audio blocks as well (tested with ogg, worked well)

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -28,6 +28,7 @@
     "@crowdsignal/http": "^0.1.0",
     "@crowdsignal/rest-api": "^0.1.0",
     "@crowdsignal/router": "^0.1.0",
+    "@wordpress/blob": "^3.2.2",
     "@wordpress/blocks": "^11.1.4",
     "@wordpress/components": "^19.1.1",
     "@wordpress/data": "^6.1.4",
--- a/apps/dashboard/src/components/editor/blocks.js
+++ b/apps/dashboard/src/components/editor/blocks.js
@@ -2,14 +2,18 @@
  * External dependencies
  */
 import { registerBlockType } from '@wordpress/blocks';
-import { addFilter } from '@wordpress/hooks';
+import { addFilter, removeFilter } from '@wordpress/hooks';
 import { forEach } from 'lodash';
 
 /**
  * Internal dependencies
  */
 import * as blocks from '@crowdsignal/block-editor';
-import { withEmptyClassName } from '@crowdsignal/block-editor-filters';
+import {
+	withEmptyClassName,
+	withMediaLinkDestination,
+} from '@crowdsignal/block-editor-filters';
+import { MediaUploader } from './media-upload';
 
 export const registerBlocks = () =>
 	forEach( blocks, ( block ) => {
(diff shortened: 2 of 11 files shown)


TheSteveK:
Just adding my contribution. 

I successfully tested and was able to upload an image via the new editor and then see it on the original editor's media library.

digitalwaveride:
Super nice Christian!

When trying here the upload seems to work at first... upload button is there... preview image appears ... 

but then the "spinner" does not disappear and I get the following error:

image

PR #146, opened 2022-01-28
Add @crowdsignal/editor-data

ice9js:
This PR adds a new `@crowdsignal/editor-data` package which extracts the editor branch from the dashboard's data store into its own, separate data store. This approach aims to clean things up a bit by drawing a hard line between the editor state and the rest of the application, which does come with some limitations.

Most notably, we're no longer able to dispatch `saveAndUpdateProject()` directly from `saveChanges()` since these actions are now in two separate stores. It is possible technically, but creates this implicit dependency on another store which may or may not exist.

To solve this problem, I added a control which relies on `crowdsignal.editor.save` action which will bind everything together, by doing something like this in the editor (which also removes the `redirect()` from the state layer):

```
addAction( 'crowdsignal.editor.save', '@crowdsignal/dashboard', ( projectId, data ) => {
    await dispatch( 'crowdsignal/dashboard' ).saveAndUpdateProject( projectId, data );

    if ( projectId ) {
        return projectId;
    }

    const id = select( 'crowdsignal/dashboard' ).getLastUpdatedProjectId();
    redirect( `/project/${ id }` );
    return id;
} );
```

It could possibly be slimmed down some more, but I just want to give an example for context.

In the state layer, I introduced a number of changes compared to the current state of `@crowdsignal/dashboard` and #141.  
`initializeEditor()` will now load the current project data into the editor state which from that point on becomes completely independent of `state.projects`, unless `initializeEditor()` is called again, in which case any unsaved changes would also be wiped out.  
So the big difference is `@crowdsignal/editor-data` always contains the complete project data, where the previous `state.editor` branch would only contain the changes if any have been made. This makes any data manipulation much easier as we do not have to do it in relation to the original data in order to be able to merge it later - an important step as our API doesn't allow for updating pages individually.  
But in order not to submit the complete project on every update which might cause unwanted side-effects, I added a `changes` reducer which tracks which parts of the projects will need to be sent depending on actions fired since the last save.

# Testing

Tests are still in WIP.

### DIFF
--- a/packages/editor-data/package.json
+++ b/packages/editor-data/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@crowdsignal/editor-data",
+  "version": "0.1.0",
+  "description": "Data layer for the Crowdsignal block editor.",
+  "author": "Automattic Inc.",
+  "license": "GPL-2.0-or-later",
+  "keywords": [
+    "crowdsignal"
+  ],
+  "homepage": "https://github.com/Automattic/crowdsignal-ui/tree/HEAD/packages/editor-data/README.md",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Automattic/crowdsignal-ui.git"
+  },
+  "bugs": {
+    "url": "https://github.com/Automattic/crowdsignal-ui/issues"
+  },
+  "main": "src/index.js",
+  "dependencies": {
+    "@crowdsignal/block-editor": "^0.1.0",
+    "@wordpress/blocks": "^11.1.4",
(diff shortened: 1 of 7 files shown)


ice9js:
Most of the changes introduced here have now been added in #141.  
As previously discussed with @CGastrell, while the code is good, moving the state to a separate package and separate `@wordpress/data` store comes with its own set of challenges which we don't want to deal with just yet.  
Especially as for the editor, there's little reward in it. The store is unlikely to be used anywhere else but the dashboard/editor anyway.

PR #147, opened 2022-02-03
Force block border style when in error state

jcheringer:
## Summary

At the current state, the block styles for the error state can be overridden by the border settings.
We need to make sure that the block error state keeps its styles no matter the user border selection.

## Test instructions

* Open or create a project and a few blocks;
* Change blocks' settings to require an answer and add some border style;
* Save and go to preview;
* Hit the submit button without filling the required blocks;
* The block's error state should override selected border settings;

### DIFF
--- a/packages/blocks/src/components/question-wrapper/index.js
+++ b/packages/blocks/src/components/question-wrapper/index.js
@@ -27,7 +27,7 @@ const StyledQuestionWrapper = styled.div`
 	}
 
 	&.is-error {
-		border-color: var( --color-error );
+		border: 1px solid var( --color-error ) !important;
 	}
 
 	${ ErrorMessage.className } {
--- a/packages/blocks/src/components/form-text-input/index.js
+++ b/packages/blocks/src/components/form-text-input/index.js
@@ -22,7 +22,7 @@ const Input = styled.input`
 	box-sizing: border-box;
 
 	${ FormInputWrapper }.is-error & {
-		border-color: var( --color-error );
+		border: 1px solid var( --color-error ) !important;
 	}
 `;
 


packages/blocks/src/components/question-wrapper/index.js
```diff
@@ -27,7 +27,7 @@ const StyledQuestionWrapper = styled.div`
 	}
 
 	&.is-error {
-		border-color: var( --color-error );
+		border: 1px solid var( --color-error ) !important;
```
ice9js, line 30: Somewhat of a nit... by fixing the border size to `1px` this can feel a little inconsistent depending on what the actual border size was.  
I wonder if it might be a better idea to use `box-shadow` for the error outline - maybe even put it on a ::before or ::after so it doesn't affect the shadow for question wrappers. Adding a pseudo element potentially also saves us the `!important` but it's not a huge deal here I think. What do you think?

ice9js (commented):
Just noted a potentially alternative way to deal with these so it doesn't mess with the border size - let me know what you think.

packages/blocks/src/components/question-wrapper/index.js
jcheringer: I think the pseudo-element is worth a try. Let me see what I can get.

CGastrell: I think we will be stuck to use `!important` here, hence, rolling back the change (use only `border-color`) and just append the `!important` directive.

One issue that comes to mind is if the user set the border color to some very similar `--color-error` color.

jcheringer: We could adjust it to use `!important` only on `border-color` and add a default `border-width` but, this would allow the user to set the `border-width` to 0 and, this way, they wouldn't see the border for the error state.

jcheringer:
As we discussed, we're going to keep the implementation like this, for now.
We can get back to this later, if necessary.

PR #148, opened 2022-02-03
Response validation error handler

CGastrell:
The PR tries to address this edgy case of someone actually skipping the frontend validations to successfully submit data that will be rejected by the backend.

Please ignore https://github.com/Automattic/crowdsignal-ui/pull/148/commits/793a6949efed6ed406c4b997837c52c7d4903067 as it's just there for the purpose of delivering un-validated values to the endpoint.

One of the goals already achieved: when the backend sends errors, we should not just *renderContent* (as we currently do), since that breaks the form functionality and also unwraps the form from its container. The introduction of `setHasErrors` on https://github.com/Automattic/crowdsignal-ui/pull/148/commits/39099c011bb5e56d739e828841f7548f3b010c11 (internal state of the project-renderer app) will provide an extra check before simply rendering the returned content.

## Backend errors
The backend is now returning the same page when an error is found and an array of error objects. Right now the errors are composed by:
- fieldClientId
- type: some sort of *const*, ie `ERROR_MANDATORY_MISSING`, `ERROR_QUESTION_SUBMIT`
- message: human readable error message. This one could be ditched and use the `type` above. It will all depend on how we handle translations

## use-validation
This PR includes a big change to how `use-validation` identifies the input. It now carries both `name` and `fieldClientId`. `name` being the required HTML attribute to use on the input and `fieldClientId` the prop used to identify the question block. These changes can be separately seen at https://github.com/Automattic/crowdsignal-ui/pull/148/commits/144192d912c1f394a89736979e91b422780776e5
Are there any downsides to this change?

## Test instructions
Checkout this branch locally and be sure to update your sandbox. Create a project with a mandatory question. Visit the project and answer without fulfilling the mandatory question. 

Errors from backend should now be rendered and form should be still functional.

### DIFF
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -40,6 +40,8 @@ const App = ( {
 
 	const [ errors, setErrors ] = useState( [] );
 
+	const formName = `f-${ projectCode }`;
+
 	useEffect( () => {
 		if ( preview ) {
 			setHostOption( 'https://api.crowdsignal.com', 'mode', 'cors' );
@@ -139,7 +141,7 @@ const App = ( {
 	return (
 		<Form
 			className="crowdsignal-forms-form"
-			name={ `f-${ projectCode }` }
+			name={ formName }
 			onSubmit={ handleSubmit }
 		>
 			<ContentWrapper className="crowdsignal-forms-form__content">
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -38,6 +38,8 @@ const App = ( {
 
 	const [ hasResponded, setHasResponded ] = useState( false );
 
+	const [ errors, setErrors ] = useState( [] );
+
 	useEffect( () => {
 		if ( preview ) {
 			setHostOption( 'https://api.crowdsignal.com', 'mode', 'cors' );
@@ -97,6 +99,7 @@ const App = ( {
 						throw new Error( 'Empty response' );
 					}
 
+					setErrors( json.errors );
 					setContent( json.content );
 					setHasResponded( true );
 					// all the setters should be called here: page, responseHash, content and startTime
@@ -129,7 +132,7 @@ const App = ( {
 			'crowdsignal-forms/text-question': TextQuestion,
 		} );
 
-	if ( hasResponded ) {
+	if ( hasResponded && errors.length === 0 ) {
 		return <ContentWrapper>{ renderContent() }</ContentWrapper>;
 	}
 
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -38,7 +38,7 @@ const App = ( {
 
 	const [ hasResponded, setHasResponded ] = useState( false );
 
-	const [ errors, setErrors ] = useState( [] );
+	const [ hasErrors, setHasErrors ] = useState( false );
 
 	const formName = `f-${ projectCode }`;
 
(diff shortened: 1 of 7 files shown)


apps/project-renderer/src/components/app/index.js
```diff
@@ -76,34 +80,32 @@ const App = ( {
...
+				}
+
+				return res.json();
+			} )
+			.then( ( json ) => {
+				if ( ! json || ! json.content ) {
+					throw new Error( 'Empty response' );
+				}
+
+				const responseWithErrors = ( json.errors || [] ).length > 0;
```
ice9js, line 103: Error responses should ideally return `400` and that's how we can detect them?
Is there anything that's preventing that? Like, the json responses aren't used anywhere else anyway right?

packages/form/src/components/form/index.js
```diff
@@ -20,11 +20,14 @@ const Form = ( { children, name, onSubmit, ...props } ) => {
...
 	);
 
-	const { startSubmit, stopSubmit } = useDispatch( STORE_NAME );
+	const { startSubmit, stopSubmit, setFieldError } = useDispatch(
+		STORE_NAME
+	);
 
-	const registerValidation = ( fieldName, validation ) =>
-		( validations[ fieldName ] = validation );
+	const registerValidation = ( fieldClientId, validation ) =>
```
ice9js, line 27: This library is meant to be context-agnostic and is not only meant to be used by blocks.  
Like, that's the case right now but we'd also be using this for more complex setting/sharing/etc. forms throughout the dashboard as we migrate those views here.

Hence `fieldName` and not `fieldClientId` - nothing stops you from using a question's field `clientId` as the value though.

It's not even clear to me what the difference is between `fieldClientId` and using `fieldName` and just maybe updating the value to make it easier to work with?

apps/project-renderer/src/components/app/index.js
```diff
@@ -76,34 +80,32 @@ const App = ( {
...
+				setHasErrors( responseWithErrors );
+				setContent( json.content );
```
ice9js, line 105: This way of doing things feels very obscure. Wouldn't it be better, if we just get `errors` ? Like:

```
{
    `errors: {
        `someClientId`: [
            `bla bla bla`,
             ...,
         ],
     },
}
```

And then we can manually dispatch `setFieldError` from here to push them straight into the form state?

apps/project-renderer/src/components/app/index.js
```diff
@@ -129,14 +131,14 @@ const App = ( {
 			'crowdsignal-forms/text-question': TextQuestion,
 		} );
 
-	if ( hasResponded ) {
+	if ( hasResponded && ! hasErrors ) {
```
ice9js, line 134: `hasResponded` shouldn't even be set until you've successfully submitted your response - I think the additional condition shouldn't be necessary?

ice9js (changes requested):
Overall, I don't understand why we don't push errors straight into the state.
Also, `@crowdsignal/form` is meant to be a general form handling library, not just for blocks.

apps/project-renderer/src/components/app/index.js
jcheringer: I'm probably guilty on this one. Christian and I had some discussions about this implementation and I suggested sending the errors to the `<Form />` and handling them from there.
Settings the field errors from the project-renderer would be much easier but I'm wondering if this is project-renderer's responsibility.
Since the `<Form /`> component already handles the fields validations, it seems more natural to me that it also handles those field errors.

ice9js: To some extend. In this case, we're not really talking about 'field errors' but rather form submission errors - which would best stay confined to the submit handler.  

To be completely honest, I still can't quite tell where the error messages are coming from or going to, or if they're used for anything at all other than setting `hasErrors`, just by reading the code.

apps/project-renderer/src/components/app/index.js
CGastrell: Agree

CGastrell:
Dropping this PR as it currently doesn't make much sense: skipping our submit process in order to send nonsense responses would effectively break the flow and wouldn't allow for further responses to be sent.

One thing that actually proved to work is exporting the STORE_NAME from the form and then dispatch `setFieldError`, we might want to come back to some of that strategy in the future.

PR #149, opened 2022-02-04
Use allowBlocks instead of disallowBlocks

ice9js:
This patch updates the editor to use a list of allowed vs. disallowed blocks. This will help us prevent any surprises if new core blocks are added - we don't necessarily want all of them to a be available in the inserter.

The list of currently allowed core blocks:

- buttons
- code
- columns
- group
- heading
- html
- image
- list
- paragraph
- preformatted
- pullquote
- quote
- row
- separator
- spacer
- table
- verse
- video

Solves c/8nwOD9BI-tr.

# Testing

Open the project editor for a new or existing projects and verify only the blocks listed above (and Crowdsignal blocks) are available in the inserter.

### DIFF
--- a/apps/dashboard/src/components/editor/settings.js
+++ b/apps/dashboard/src/components/editor/settings.js
@@ -18,14 +18,31 @@ setCategories( [
 export const editorSettings = {
 	iso: {
 		blocks: {
-			disallowBlocks: [
-				'core/audio',
-				'core/calendar',
-				'core/cover',
-				'core/embed',
-				'core/file',
-				'core/media-text',
-				'core/shortcode',
+			allowBlocks: [
+				'core/buttons',
+				'core/code',
+				'core/columns',
+				'core/group',
+				'core/heading',
+				'core/html',
+				'core/image',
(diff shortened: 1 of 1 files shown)


jcheringer (approved):
LGTM!

digitalwaveride (changes requested):
Left-sidebar library is filtered! Awesome!


### Search still 

Via search... e.g. in the inserter in the editor, I have still access to all the blocks... including not approved ones:

image


### Quick Block List Update

We should put the "Cover Block" on our list now... as we support the upload now! Great block.

We likely can hide the "verse" block, probably will not see high usage in Crowdsignal context.

CGastrell (approved):
This is as far as we can get with this, the searcher clearly doesn't honor the disallowed/allowed block list. I'd just add what Thomas suggested and move on with this.

digitalwaveride:
Happy to move on with this for now! 👍 

... but we will need to come back to this soonish, unfortunately... 😞 - block exploration will be a core onboarding challenge.

PR #150, opened 2022-02-04
Update MC Question block sidebar behavior

jcheringer:
## Summary

The purpose of this PR is to improve the sidebar behavior of the MC Question block when interacting with the fields: `An answer is required`, `Min. Choinces`, `Max Choices`.
Interacting with those fields will change the MC block attributes to be consistent with the selection.

Examples:

* Changing the `Max Choices` field to >= 2 will turn the block into a `Multiple Choice` question;
* Changing the `Max Choices` to 1 will make it a `Single Choice` question;
* `Max Choices` can't be less than 1;
* `Max Choices` can't be lower than `Min Choices`;
* `Min Choices` can't be greater than `Max Choices`
* `Min Choices` can't be less than 0;
* Changing `Min Choices` to 0 will make the question not required;

## Test Instructions

* Open or create a new project and add an MC Question block;
* Play with the `Required`, `Min Choices` and `Max Choices` fields and check if they are consistent;
* You can use the rules described above as a starting point;

### DIFF
--- a/packages/block-editor/src/multiple-choice-question/sidebar.js
+++ b/packages/block-editor/src/multiple-choice-question/sidebar.js
@@ -12,15 +12,57 @@ import BorderSettings from '../components/border-settings';
 import ColorSettings from '../components/color-settings';
 
 const Sidebar = ( { attributes, setAttributes } ) => {
-	const handleChangeAttribute = ( key ) => ( value ) =>
+	const handleChangeMandatory = ( isMandatory ) => {
 		setAttributes( {
-			[ key ]: value,
+			mandatory: isMandatory,
 		} );
 
-	const handleChangeNumberAttribute = ( key ) => ( value ) =>
+		if ( isMandatory && attributes.minimumChoices < 1 ) {
+			setAttributes( {
+				minimumChoices: 1,
+			} );
+		} else if ( ! isMandatory ) {
+			setAttributes( {
+				minimumChoices: 0,
+			} );
+		}
+	};
+
+	const handleMinChoices = ( value ) => {
+		if ( value < 0 ) {
+			return;
+		}
+
 		setAttributes( {
-			[ key ]: parseInt( value, 10 ),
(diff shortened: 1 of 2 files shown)


digitalwaveride (changes requested):
Awesome Jõao!

Thanks a ton for tackling this 🙏 

### In "answer is **not** required state" - can't delete value input field for max choices

If there is a value in the max choice field, and I want to change it from a 3 to a 2, I can't delete the 3, as it seems to not allow an empty input field. That behaviour seems to be expected though by users when editing an input field with the goal to change a value ... 
In this case I can edit the field only via the arrows, but not via keyboard... 

image


### In required state: Hiding the "min choices" field

As discussed on Trello, we should hide for now the min choice field for simplicity reasons.
This will be a  deliberate decision, very well being aware that we don't serve some edgy use cases like "please choose at least two but not more than 5 choices".

For a couple of reasons... 
-> Most likely the survey question should not be designed like that, as such complex answering rules are confusing to the respondent, and lead to a bad experience.
-> It's likely an edge with less than 5% usage and simplifying the complex Gutenberg UI is a top priority for us.
-> Even if a user is missing that feature, they should be ok with no "min option" for now as well
-> Failure message handling for users would need to be much more precise on our end (one still missing... two too much... )
-> We would likely need to introduce error messages inside of the right editor sidebar as well, as the involved logic in the current implementation would do too many surprising things in the UX ...like jumping between required/unrequired state with hiding the input field when deleting the min choice input field to add a new value, not being able to properly edit the max choice label, if a min choice label is set ... 

We should go for a reduced complexity here!
If users really need that feature, they will tell us, and we can add it at a later point.

jcheringer:
Hey @digitalwaveride! I did some adjustments, could you take a look again, please? 🙏 
Thanks!

digitalwaveride (approved):
Nice João!
Thanks so much! Worked 🚀 

Nice touch, disabling the "0" ;)

PR #152, opened 2022-02-08
Add parseTagProps for exceptional treatment on props by tag

CGastrell:
This PR adds a props parser amidst the rendering process.

This allows for place to give some props special treatment. The particular case in this PR being the `<audio>` tag, which requires a `controls` prop to be passed on or it would not render at all (by design).

The `core/audio` block does add this `controls` prop, but as a boolean prop it is assumed `true` by its mere presence on HTML markup. This doesn't play nice with React's createElement or other parsing mechanisms as the prop arrives as `controls: ''` effectively evaluating to false or not being parsed at all.

For this and other cases, we can start adding exceptions via `parseTagProps`.

## Test instructions
Enable `core/audio` by commenting the entry on `apps/dashboard/src/components/editor/settings.js` line 22:
```
// 'core/audio'
```

Use the editor to create an audio block, upload a wav/mp3/ogg from your computer. Visit the public view of the project, you'll notice the audio block is not there (yet if you inspect the page, you'll see the markup without the `controls` prop)

Checkout `add/tag-props-parser` and visit the public view of the  project again. This time you should see the audio block.

### DIFF
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -62,7 +62,7 @@ const parseNodes = ( nodes ) =>
 			name: toLower( node.tagName ),
 			children: parseNodes( node.childNodes ),
 			props: {
-				...parseTagProps( toLower( node.tagName ), attributes ),
+				...parseTagProps( node.tagName, attributes ),
 				className,
 				style: parseStyles( style ),
 				key: uniqueId( KEY_PREFIX ),
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -56,8 +56,9 @@ const parseTagProps = ( tagName, props ) => {
 		) {
 			props.controls = 'controls';
 		}
-		return props;
 	}
+
+	return props;
 };
 
 /**
--- a/packages/blocks/src/render-blocks.js
+++ b/packages/blocks/src/render-blocks.js
@@ -44,6 +44,25 @@ const parseStyles = ( styleString ) => {
 /**
  * Parses DOM nodes into an array of objects to be used to create React elements.
  *
+ * @param  {string}  tagName HTML tag
+ * @param  {Object}  props object
+ * @return {Object}  Props object
+ */
+const parseTagProps = ( tagName, props ) => {
+	if ( tagName === 'AUDIO' ) {
+		if (
+			typeof props.controls !== 'undefined' &&
+			props.controls !== false
+		) {
+			props.controls = 'controls';
+		}
+		return props;
+	}
+};
+
+/**
+ * Parses DOM nodes into an array of objects to be used to create React elements.
+ *
  * @param  {HTMLCollection} nodes
  * @return {Array}                Array of objects describing the elements to be created.
  */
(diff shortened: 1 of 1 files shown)


CGastrell:
Wondering what you think about this one @ice9js

packages/blocks/src/render-blocks.js
```diff
@@ -62,14 +62,34 @@ const parseNodes = ( nodes ) =>
 			name: toLower( node.tagName ),
 			children: parseNodes( node.childNodes ),
 			props: {
-				...attributes,
+				...parseTagProps( toLower( node.tagName ), attributes ),
```
ice9js, line 65: Don't really need `toLower` here - `node.tagName` is guaranteed to be all caps here and it's hard to even say how much of a convenience it is to have it in lowercase in the switch statement below since the value isn't used anywhere else?

packages/blocks/src/render-blocks.js
```diff
@@ -62,14 +62,34 @@ const parseNodes = ( nodes ) =>
...
 	} );
 
+/**
+ * Parses DOM nodes into an array of objects to be used to create React elements.
+ *
+ * @param  {string} tagName
+ * @param {Object}  props object.
+ * @return {Object}  Props object.
+ */
+const parseTagProps = ( tagName, props ) => {
```
ice9js, line 80: I'm actually surprised this works with an arrow function defined below where it's first used.  
Let's maybe still put it on top to keep the dependency flow.

packages/blocks/src/render-blocks.js
```diff
@@ -62,14 +62,34 @@ const parseNodes = ( nodes ) =>
...
 				key: uniqueId( KEY_PREFIX ),
 			},
 		};
 	} );
 
+/**
+ * Parses DOM nodes into an array of objects to be used to create React elements.
+ *
+ * @param  {string} tagName
+ * @param {Object}  props object.
```
ice9js, line 77: Nit: can we align these? I actually thought our linter should be picky about that 🤔

packages/blocks/src/render-blocks.js
```diff
@@ -62,14 +62,34 @@ const parseNodes = ( nodes ) =>
...
+ */
+const parseTagProps = ( tagName, props ) => {
+	switch ( tagName ) {
+		case 'audio':
+			if (
+				typeof props.controls !== 'undefined' &&
+				props.controls !== false
+			) {
+				props.controls = 'controls';
+			}
```
ice9js, line 88: `break;` statement is missing.  

Personally, I'd prefer to get rid of the `switch` altogether and replace it with simple `if` statements and early returns.

ice9js (changes requested):
I think it makes sense - I just left a few minor comments before we merge it.

packages/blocks/src/render-blocks.js
CGastrell: I changed this into a simple if, yet with its current use there's no point for an early return. Eventually other exceptions might need to be added but, for now, mutating the props and return at the end should suffice. Remember this function needs to return unmodified props when no tagName is matched

ice9js (approved):
:shipit:

PR #153, opened 2022-02-10
Add page navigation drag&drop

ice9js:
This patch adds drag&drop functionality to page navigation.

~Unfortunately, there's a small caveat in that the entire 'page' preview is a button and already has a function (selecting a page), hence it's not possible for to serve as a drag&drop handle as well.  
In order for it to work, I added a small handle that appears on the left when you hover over a page, but we'll likely need to come up with a more streamlined way to integrate that. cc @digitalwaveride~

Solves c/rkfdI2Dm-tr.

# Testing

- Open a project with multiple pages or create one.
- You should be able to drag and drop different pages and reorder them that way.
- Reordering pages should trigger autosave.
- Refresh the page and make sure the reordered content has been saved correctly.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -41,6 +41,7 @@
     "classnames": "^2.2.5",
     "isolated-block-editor": "git://github.com/Automattic/isolated-block-editor.git#2.9.0",
     "lodash": "^4.17.21",
+    "react-beautiful-dnd": "^13.1.0",
     "react-helmet": "^6.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


digitalwaveride:
Awesome Kuba! Things are starting to make more and more fun!
Thanks so much for building this!

Based on our chat from yesterday evening:

In a best-case scenario, the drag & drop could be activated on a "long click" anywhere on the preview window row.

If that is not currently doable, a drag handle on the left would be a compromise. we should try though to maximize the active area to drag. Compared to the version here, it would be great to expand the area besides the handle, also including the "number".


image


For optimizing the space, a bit of fine-tuning, moving the preview screen more to the right, making more space for the standard Gutenberg "handle" on the left. 


image


Here is a rough prototype for what the interaction could look like:




https://user-images.githubusercontent.com/5120357/153576436-b972ebd8-2aa7-43ac-9219-d911d18cb54d.mov

ice9js:
Seems to work now with both drag and drop and click behavior now - just needed to do some more digging to find the appropriate settings.

digitalwaveride (approved):
So awesome Kuba!!! 😍 

We still might want to add the drag handler? ... but not a blocker right now!

PR #154, opened 2022-02-11
Implement a toggle animation for the page navigation column

ice9js:
This PR implements the toggle and animation for triggering the page navigation column. The collapsed state only features the page numbers.

Collapsed:

Screen Shot 2022-02-11 at 10 42 53 PM

Expanded:

Screen Shot 2022-02-11 at 10 36 37 PM

Solves c/N1eM8RnL-tr.

# Testing

Clicking the 'Pages' toggle on top of the page navigation column will should collapse/expand it depending on current state.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -43,6 +43,7 @@
     "lodash": "^4.17.21",
     "react-beautiful-dnd": "^13.1.0",
     "react-helmet": "^6.1.0",
+    "react-transition-group": "^4.4.2",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


digitalwaveride (approved):
🚀 🚀 🚀

CGastrell (approved):
Sweet! I never fully understood how the transition group works, but it looks awesome!

ice9js:
>Sweet! I never fully understood how the transition group works, but it looks awesome!

It's just a fancy timer really, and gives you slightly more control about when animations should start - so you can for example still trigger animations when mounting elements.  

While we're at it, it's worth noting the previews are only hidden and never actually unmount. I tried it but the performance is actually quite terrible when they all try to mount at once when you expand the sidebar.

PR #155, opened 2022-02-14
Include confirmation page in editor page navigation

ice9js:
This PR adds the capability to edit the project confirmation page in the block editor. The page sits in its own dedicated 'confirmation' section of the page navigation sidebar - like discussed with @digitalwaveride.

The way it's implemented, we treat the last page of the project's content as its confirmation page. This means from now on, every new project includes two pages by default.  
The confirmation page cannot be deleted, reordered and generally doesn't offer any of the functions regular pages might have.

crowdsignal localhost_9000_project

Solves c/8Svmd41d-tr.

# Testing

- Create a new project, it should contain a blank page as well as a confirmation page with default text.
- Question, input and submit button blocks should be unavailable when editing the confirmation page.
- It shouldn't be possible to reorder, swap or delete the confirmation page.
- The confirmation page shouldn't be taken into account when checking if every page contains at least one submit button.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -42,7 +42,8 @@
     "isolated-block-editor": "git://github.com/Automattic/isolated-block-editor.git#2.9.0",
     "lodash": "^4.17.21",
     "react-beautiful-dnd": "^13.1.0",
-    "react-helmet": "^6.1.0"
+    "react-helmet": "^6.1.0",
+    "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@babel/plugin-proposal-optional-chaining": "^7.16.7",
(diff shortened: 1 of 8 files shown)


digitalwaveride:
Nice Kuba! Great to have the "Thank You Page in the game now! 😃 

Two things I noticed!

- When I add a new page, the "Thank You" page gets pushed up to be page 2, the new page is the "confirmation" page:

image


- I was able to add submit buttons on all the pages... 

image

digitalwaveride (approved):
Works now as advertised! Thanks Kuba! 🚀

CGastrell (approved):
Works nicely! The only misbehave I saw was this glitch: while at the confirmation page, adding a new page will set that new page as active, yet you'll still see the confirmation page on the editor. Clicking on the already active (new) page will do nothing. You have to click on another page and back to the new one to make it work. As said, just a minor glitch.

PR #156, opened 2022-02-14
Add Embed block support

jcheringer:
## Summary

The purpose of this PR is to add support for the WP Embed Block.

Depends on: D74885-code
Ref: c/ellkRtk8-tr

## Test Instructions

* Open o create a project
* Paste a URL from one of the supported providers or manually add the Embed Block and update the URL field
* The embed content should be loaded correctly


digitalwaveride:
This is wild! Thanks so much João! Really loved to play with this:

Inside of the editor, embed block worked for me, just a little glitch with our very own poll block, where the block height was off... 

image


There is though an issue on the Preview side of things... 

Only the YouTube block got properly rendered, a Soundcloud or Crowdsignal poll block did not get rendered:

image

CGastrell:
I have little to no saying here, work is solid and looks good. Tests through different providers should get us on a selection of "working" embeds and the rest should be addressed as we go (being our own embeds something to prioritize).

That said, a couple of thoughts:
- as in the diff, I assume this is the first step towards a fully working embed block and further security/consistency checks are needed
- why the `core/embed` name? It might get confusing

jcheringer:
> This is wild! Thanks so much João! Really loved to play with this:
> 
> Inside of the editor, embed block worked for me, just a little glitch with our very own poll block, where the block height was off...
> 
> There is though an issue on the Preview side of things...
> 
> Only the YouTube block got properly rendered, a Soundcloud or Crowdsignal poll block did not get rendered:

Hey Thomas! Thanks for testing it.
I was able to fix the issue on the editor but I can't say the same for the project-renderer.
It might be something I'm missing but, after a few hours of debugging, I wasn't able to figure it out.
Do you think it's worth having this PR merged a handle the project-renderer issue on another PR?

jcheringer:
Hey @CGastrell!

> * as in the diff, I assume this is the first step towards a fully working embed block and further security/consistency checks are needed

Well, the embed block implementation on `crowdsignal-ui` was pretty much based on the [embed-preview.js](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/embed/embed-preview.js). ATM, it's not clear to me what security/consistency checks we might need to add but, for sure we can do that as we go.

> * why the `core/embed` name? It might get confusing
I tried to keep it consistent with the block name we get from the DB.
Since we're overriding the block we want to show on the project-renderer, we need to [map core/embed](https://github.com/Automattic/crowdsignal-ui/pull/156/files#diff-a1a9627a4c8a40a4a0cb7f025026f2fe11ab3aeda3d53a4e361a336439f3f8b5R126) to a block on our side.

CGastrell (approved):
This still needs work, but it's good enough up to this point. I'd merge it and start looking at how we want it to behave on special cases.

digitalwaveride:
👍 As discussed in our call today, let's get it merged and start debugging from there!

PR #157, opened 2022-02-15
Set internal state when response comes back

CGastrell:
This PR addresses some needed changes on the project renderer to properly submit pages to backend.

Ref: D75039-code

## Test instructions
Checkout and apply corresponding backend patch, follow the instructions there.

### DIFF
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -101,7 +101,7 @@ const App = ( {
 					setContent( json.content );
 					setHasResponded( json.done );
 					setResponseHash( json.r );
-					setCurrentPage( parseInt( json.p ) );
+					setCurrentPage( parseInt( json.p, 10 ) );
 				} )
 				// eslint-disable-next-line no-console
 				.catch( ( err ) => console.error( err ) )
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -99,8 +99,9 @@ const App = ( {
 					}
 
 					setContent( json.content );
-					setHasResponded( true );
-					// all the setters should be called here: page, responseHash, content and startTime
+					setHasResponded( json.done );
+					setResponseHash( json.r );
+					setCurrentPage( parseInt( json.p ) );
 				} )
 				// eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


digitalwaveride:
Thanks Christian for tackling this!!! 🙏 

Two things that were off in my setup:

- Only changes on page 1 were updating/saved, any changes on page 2 were not updated, also the "update button" seemed to behave weirdly as it always stayed active, even though no changes were done.

https://user-images.githubusercontent.com/5120357/154363894-737d12ce-b9c2-4d4e-bed8-3b682b0ca18c.mov


----
- After clicking submit on the second page, it did not show the "confirmation page".
Responses for the MC question on page 2 were not counted:

image

CGastrell:
Yeah, I've seen the editor behaving strangely, but this has nothing to do with this change. We were talking about this with @ice9js and there seems to be some regression issues we'll address later. For the purpose of this change, be sure to have a multiple page project, then visit its public view at `https://crowdsignal.localhost:9001/[projectCodeHere]` (also running from `yarn start`).

If you're struggling to publish the project, make a change and BEFORE the autosave triggers, hit the "Publish" button (this is the regression I was talking about above).

apps/project-renderer/src/components/app/index.js
```diff
@@ -98,8 +98,9 @@ const App = ( {
 					}
 
 					setContent( json.content );
-					setHasResponded( true );
-					// all the setters should be called here: page, responseHash, content and startTime
+					setHasResponded( json.done );
+					setResponseHash( json.r );
+					setCurrentPage( parseInt( json.p ) );
```
ice9js, line 103: Make sure to add the base to `parseInt` -> `parseInt( json.p, 10 )` [reference](https://stackoverflow.com/questions/6611824/why-do-we-need-to-use-radix-parameter-when-calling-parseint).
Also, is there any change we can just have it be a number in the first place?

ice9js (commented):
Just one minor issue but should be good to go afterwards.

apps/project-renderer/src/components/app/index.js
CGastrell: The cryptic response props are there to honor our legacy responses' data structure. The `p` stands for "page" and it's sent as an `int`, but I though we'd just make sure of it on the frontend.

PR #158, opened 2022-02-17
Remove @wordpress/block-editor dependency from @crowdsignal/blocks

ice9js:
This patch replaces all `RichText.Content` usages inside `@crowdsignal/blocks` with `RawHTML` which allows us to remove the dependency on `@wordpress/block-editor`.

Removing this dependency results in `main.js` shrinking from 1.1MB to 198kB when built for production.

# Testing

Verify our question, input and submit button blocks still display text correctly.
Pay special attention to special characters, formatting options and emojis.

### DIFF
--- a/packages/blocks/src/core-embed/index.js
+++ b/packages/blocks/src/core-embed/index.js
@@ -12,7 +12,7 @@ import classnames from 'classnames';
 import { Sandbox } from '@crowdsignal/components';
 import { fetchEmbedContent } from '@crowdsignal/rest-api';
 
-const getPhotoHtml = ( photo ) => {
+const getPhotoHTML = ( photo ) => {
 	const imageUrl = photo.thumbnail_url || photo.url;
 	const photoPreview = (
 		<p>
@@ -41,7 +41,7 @@ const CoreEmbed = ( { attributes } ) => {
 		'is-type-video': 'video' === type,
 	} );
 
-	const html = 'photo' === type ? getPhotoHtml( preview ) : preview.html;
+	const html = 'photo' === type ? getPhotoHTML( preview ) : preview.html;
 
 	const iframeTitle = sprintf(
 		// translators: %s: host providing embed content e.g: www.youtube.com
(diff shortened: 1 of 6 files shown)


ice9js:
Seems #156 introduced additional dependencies on `@wordpress/components` and `@wordpress/core-data` which blow the bundle back up to 1.3MB and will need to be cleaned up too.

CGastrell (approved):
Seems to cope fine with emojis and characters:
image

image

The most offending difference is our submit button style, which has nothing to do with this PR or anything

CGastrell:
Tags seem to be doing ok as well
image

ice9js:
Rebased & refactored our `core-embed` implementation to remove any `@wordpress/*` dependencies other than `@wordpress/element`.

packages/blocks/src/core-embed/index.js
```diff
@@ -22,20 +25,17 @@ const getPhotoHtml = ( photo ) => {
...
+	const [ preview, setPreview ] = useState( {} );
 
-			return {
-				preview: validPreview ? embedPreview : {},
-			};
-		},
-		[ url ]
-	);
+	useEffect( () => {
+		fetchEmbedContent( { url } ).then( ( { data } ) => {
```
jcheringer, line 31: On WP, there is a piece of code that removes a wrapper div from the response. Ex `<div class="embed-youtube">`.
This is necessary to prevent some styling issues.
On this PR, if you embed a YouTube video and take a look at it on the Project Renderer, it'll be cropped.
When I first implemented the embed block support, I added this piece of code inside the [embedRequestInterceptor](https://github.com/Automattic/crowdsignal-ui/blob/main/packages/rest-api/src/embed/index.js#L14), since we were using the WP components, both requests, for the Editor and for the Project Renderer were intercepted and handled properly.
Now that we're using `fetchEmbedContent` directly, we'll need to adapt the code to remove the wrapper div.

jcheringer (changes requested):
Looks good! Thanks @ice9js!
Just added a comment about a styling issue with the Embed block on the Project Renderer.

jcheringer (approved):
Awesome! LGTM! :shipit:

PR #159, opened 2022-02-18
Add page preview context menu and a duplicate option

ice9js:
This patch adds an expandable context menu to page previews inside the editor, replacing the existing delete page button.  
The new menu offers an additional _'Duplicate'_ options which allows duplicating pages.

It's worth noting that since pages live 'above' the editor - our `useClientId()` hook wouldn't kick in until you visit both the original and the duplicated page. Hence it's necessary to manually regenerate all `attributes.clientId` values for our blocks whenever a page is being duplicated.  
This isn't an issue for Gutenberg's block `clientId`s as they're only used internally by the editor itself and sort themselves out anytime the editor is reloaded.

Screen Shot 2022-02-18 at 1 28 24 PM

Solves c/X11AH7af-tr.

# Testing

- Create a new project and proceed to add some content.
- Hovering over a page preview should display a button to trigger the new context menu.
- Try duplicating the page, verify that when the draft is saved any question on the duplicated pages have unique `attributes.clientId` values.
- Try deleting a page, it should work the same way as previously.
- The context menu isn't available for the confirmation page which can't be duplicated or deleted.


CGastrell (approved):
Works nicely! One little issue with the popup menu: once I clicked "Delete", the menu remains.

Apparently the menu kept open retains functionality so clicking "Delete" keeps attempting to delete the page number you originally intended. So, delete 3rd page, click "Delete" again and the "new 3rd page" will be deleted. So on, so on... until you reach the last one: at this point the menu disappears (right on time so we don't delete the "Confirmation page".

I'd say let's take care of this on a follow up PR.

digitalwaveride (approved):
Looks great Kuba! 🚀

ice9js:
>Apparently the menu stays open and retains functionality: clicking "Delete" keeps deleting the page number you originally intended. So, delete 3rd page, click "Delete" again and the "new 3rd page" will be deleted. So on, so on... until you reach the last one: at this point the menu disappears, right on time to avoid deleting the "Confirmation page".

Actually, we still need a confirmation dialog for the page delete button (no undo) so this should be solved by that as well.

PR #160, opened 2022-02-18
Add placeholder to Text Question and Text Input blocks

jcheringer:
## Summary

The purpose of this PR is to include a `placeholder` option on the Sidebar for the Text Question and Text Input block.

Ref: c/hKFZL7jl-tr

## Test Instructions
* Add a Text Question and a Text Input block
* On the Sidebar of each block, you should see a new Placeholder option
* By setting this value, you should see it on both, the Editor and the Project Renderer

### DIFF
--- a/packages/block-editor/src/text-input/attributes.js
+++ b/packages/block-editor/src/text-input/attributes.js
@@ -7,6 +7,10 @@ export default {
 		type: 'string',
 		default: '',
 	},
+	placeholder: {
+		type: 'string',
+		default: '',
+	},
 	mandatory: {
 		type: 'boolean',
 		default: false,
--- a/packages/block-editor/src/text-input/edit.js
+++ b/packages/block-editor/src/text-input/edit.js
@@ -63,7 +63,10 @@ const EditTextInput = ( props ) => {
 					height: `${ attributes.inputHeight }px`,
 				} }
 			>
-				<FormTextInput.Preview className="crowdsignal-forms-text-input-block__wrapper" />
+				<FormTextInput.Preview
+					placeholder={ attributes.placeholder }
+					className="crowdsignal-forms-text-input-block__wrapper"
+				/>
 			</ResizableBox>
 		</FormInputWrapper>
 	);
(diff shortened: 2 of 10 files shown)


digitalwaveride (changes requested):
Thanks João for going after this!!! 🙏 

We should try to go as far as possible with Gutenberg, as a "sitebuilder", to a WYSIWYG experience .

In the case of a placeholder text this should be doable inside of the editor, visible and easy to manipulate by the user, without any sidebar at all (in order to reduce noise there as well)

The plan would be to have an implementation similar to the Feedback Button and NPS block.



Feedback Button Block

image


NPS Block 

image

digitalwaveride:
One additional comment to the placeholder behavior: 

There is actually a difference to the Feedback / NPS block example.

In the case of these two placeholders here, we don't really know what the default text of these questions are, so we have a hard time to make a good suggestion, that works by default if the user ignores it.

Here is a possible solution:

- The placeholder text says by default: "Enter placeholder"
- If this default placeholder text is **not** edited by the user, we ignore it in the preview and published view, input field will be empty.
- If the user edits the placeholder text, we display the users version as a placeholder text.

image




And a tiny nitpicking on the design:

The text input form placeholder text in the editor view is not aligned regarding line-height - compared to preview view, it's a tiny bit too high. 

image

jcheringer:
Hey @digitalwaveride! Thanks for taking a look at this.
I Did some changes trying to mimic what we have on the Feedback button.
Could you test it again, please?
Thanks! 🙏

CGastrell (approved):
LGTM! :shipit:

digitalwaveride:
Thanks so much João!!! 💪 🚀 

Awesome!

On little pixel pushing thing:

In the editor view for the Text Input Form block, the placeholder text is not completely centered horizontally... much better in the Preview view:

Editor View

image


Preview View

image

jcheringer:
I had forgotten the alignment issue, sorry @digitalwaverie.
This seemed to be an issue only on Firefox, on Chrome it looked fine.
But there was a font-family difference, though.
I did an adjustment to make the Editor and the Project Renderer use the same font-family and it seems to be good now.
Could you confirm, please?
Thanks a lot!

digitalwaveride (approved):
Fixed! Looks good! Thanks a ton, João!!! 🚀

PR #162, opened 2022-02-22
Fix publish button

ice9js:
This patch fixes project publishing which was suffering from a regression causing it not to do anything if all changes have already been saved previously.

I did this by bumping up the logic for appending the public data from the action generator to `getEditorUpdatedProjectData` which I guess makes sense considering what it's used for/how its named.

Solves c/sZ0d1Kct/-tr.

# Testing

- Create a project, wait for all changes to save, hit 'publish'. Your project should be published correctly.
- Open an existing draft project, hit 'publish'. Your project should be published correctly.


CGastrell (approved):
Thanks @ice9js Works nicely now! :shipit:

PR #163, opened 2022-02-22
Add submit page for previews

CGastrell:
This PR implements preview submit capabilities for the project-renderer.

A new method has been added to the rest-api package: submitProjectform

NOTE: an issue became evident while working on this PR where the form would submit the entire project data every time. This is being addressed on #164 

Depends on D75489-code

## Test instructions
Create a project with a couple of pages. Draft or Publish it and proceed to open the preview.
You should be able to submit all pages without restriction as long as you are the owner

### DIFF
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -19,7 +19,7 @@ import {
 import { Form } from '@crowdsignal/form';
 import { useStylesheet } from '@crowdsignal/hooks';
 import { setHostOption } from '@crowdsignal/http';
-import { fetchProjectForm } from '@crowdsignal/rest-api';
+import { fetchProjectForm, submitProjectForm } from '@crowdsignal/rest-api';
 
 const App = ( {
 	projectCode,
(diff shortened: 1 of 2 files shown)


apps/project-renderer/src/components/app/index.js
```diff
@@ -78,22 +87,12 @@ const App = ( {
...
-						throw new Error( res.status );
-					}
-
-					return res.json();
-				} )
-				.then( ( json ) => {
+			submitProjectForm(
+				projectCode,
+				form,
+				preview ? { preview: true } : {}
```
jcheringer, line 93: Couldn't we just pass `{ preview }` to the query param?
IMO, I think it'd make the code a bit cleaner avoiding the ternary if condition.
I'm assuming we're validating the value of `preview` on the backend to check whether it's true or false.

apps/project-renderer/src/components/app/index.js
```diff
@@ -69,6 +69,15 @@ const App = ( {
 			data = {};
 		}
 
+		if ( preview ) {
```
ice9js, line 72: Actually, on a second look - this shouldn't be necessary at all. It's already enabled in `useEffect` above if `?preview` is set.

ice9js (changes requested):

apps/project-renderer/src/components/app/index.js
CGastrell: That should be true, let me test that

apps/project-renderer/src/components/app/index.js
CGastrell: Well... I do agree the code would look better. Then again, `preview` is either `true` or `undefined` and when the JS-Request-PHP translation happens, sending a `{ preview }` as querystring ends up, at PHP's side, as a `(string) "undefined"` (which is not a *falsy* value).

So, I guess it's just a matter of where do we put the ugly code :)

jcheringer (approved):
LGTM! 🚢

PR #164, opened 2022-02-22
Reset form state after submit

CGastrell:
This PR adds a reset action/reducer to our `form` package. It also adds a call to the reset after the form has been submitted.

This addresses an issue that became evident while working on #163 
While the problem is not easily spotted on our tests, a large project could pile up a huge payload if we submit the entire project values each time. Besides, this is prone to errors as each page should only be submitting its own fields.

## Test instructions
Create a project with a couple of pages and inputs in it. Use the devtools to keep an eye on the requests to backend.
After submitting each page, the request payload should only carry the fields (and values) present on each page.

@ice9js @jcheringer I can't be sure this "reset" doesn't mess with other ideas we might have had for the form. I'm also trusting on the promise chain to skip the form's `onSubmit.then` in case of errors (otherwise, the form will reset itself on any errors, effectively emptying the current page values)

### DIFF
--- a/packages/form/src/components/form/index.js
+++ b/packages/form/src/components/form/index.js
@@ -20,7 +20,7 @@ const Form = ( { children, name, onSubmit, ...props } ) => {
 		[ name ]
 	);
 
-	const { startSubmit, stopSubmit } = useDispatch( STORE_NAME );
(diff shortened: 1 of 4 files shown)


ice9js (approved):
👍  🙇‍♂️  I approve.

PR #165, opened 2022-02-25
Add a new project wizard dialog

ice9js:
This patch adds a new dialog component to be displayed when starting new project. As of right now, the dialog only allows for selecting a template but will be expanded in the future.

It's worth noting, displaying the dialog in front of the editor is actually of importance as that ensures all blocks and the theme are loaded. While it is possible to render previews, it's actually a great deal of effort to make that happen.

Note that the templates themselves are yet to be fine tuned in terms of their structure and we'll possibly want to move them out to their own package - like `@crowdsignal/project-templates` - eventually.  
`attributes.clientId` values **must** be generated for all `crowdsignal-forms/` blocks before the template is inserted into the editor. `useClientId()` only works if the page becomes active, so if somebody were to choose a template and immediately publish, it's possible our blocks would be missing client IDs otherwise.

select-template

Solves c/9cycqiaH-tr.

# Testing

- Go to `/project` - you should see the new project wizard.
- Clicking on a template should auto-populate your new project with that template.
- Selecting a template **should not** trigger an autosave, but it **should** make the editor go into unsaved changes state.
- When opening an existing project, the project wizard should not appear.

### DIFF
--- a/apps/dashboard/src/components/app/styles.js
+++ b/apps/dashboard/src/components/app/styles.js
@@ -5,6 +5,18 @@ import styled from '@emotion/styled';
 import { css } from '@emotion/core';
 
 export const pageStyles = css`
+	@font-face {
+		font-family: Recoleta;
+		font-weight: normal;
+		src: url( 'https://crowdsignal.com/wp-content/themes/a8c/crowd-signal/assets/fonts/recoleta-regular-webfont.woff' );
+	}
+
+	@font-face {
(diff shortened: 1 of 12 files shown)


CGastrell (approved):
Works nicely! :shipit:

digitalwaveride:
👍 Looking good! Works for me!

Thanks Kuba! 🚀

PR #166, opened 2022-02-28
Add 'finish resolution' action

CGastrell:
This PR adds FINISH_RESOLUTION action type/creator to actively, well, finish the resolve resolution and avoid a full re-render of the editor.

NOTE: I found no way to import the action creator from the `@wordpress/data` package (no specific export, full path: https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/redux-store/metadata/actions.js), if there's a way, I'd rather have it imported instead of duped.

## Test instructions
Create a new project. See that upon first save the editor doesn't do a full reload (it should be evident if it does as the loader takes around 2 seconds when this happens).
Publish the project. The PublishButton, Share and Preview should respond as expected


apps/dashboard/src/data/projects/actions.js
```diff
@@ -59,6 +59,10 @@ export function* saveAndUpdateProject( projectId, project ) {
 			? yield dispatchAsync( patchProject, [ projectId, project ] )
 			: yield dispatchAsync( createProject, [ project ] );
 
+		yield finishResolution( 'getProject', [
+			response.data.id ? String( response.data.id ) : '0',
```
ice9js, line 63: Is it possible for `response.data.id` to not be a number here?

apps/dashboard/src/data/actions.js
```diff
@@ -1,10 +1,16 @@
...
-import { ASYNC_ROUTINE_DISPATCH } from './action-types';
+import { ASYNC_ROUTINE_DISPATCH, FINISH_RESOLUTION } from './action-types';
 
 export const dispatchAsync = ( apply, args = [] ) => ( {
 	type: ASYNC_ROUTINE_DISPATCH,
 	apply,
 	args,
 } );
+
+export const finishResolution = ( selectorName, args ) => ( {
```
ice9js, line 12: It might be worth adding a docblock comment here with a link to the original action or something - basically explaining what this is and why it's here.

ice9js (approved):
Nice, thanks Christian! I left two minor comments but good to go otherwise.

PR #167, opened 2022-02-28
Attempt to cast int on number-ish param match

CGastrell:
This PR adds an attempt to cast to integer any matched param that `isNumber`. Props to @ice9js coming up with this nit while we were discussing possible solutions on #166 

No actual testing besides logging what's going on Route.


ice9js (approved):
:shipit:

PR #168, opened 2022-02-28
Fix missing editor settings key crashing the editor

ice9js:
This patch fixes a small regression introduced in #149 where a change to `editorSettings` structure has caused the editor to crash due to other changes that have been added since.

# Testing

The editor should no longer crash when you try to go to `/project` or `/project/:projectId`.


CGastrell (approved):
Works! Thanks! :shipit:

PR #169, opened 2022-03-02
Add a page delete confirmation dialog

ice9js:
This patch adds a simple confirmation dialog for deleting pages from a project.

# Testing

Try to delete a page - a confirmation dialog should appear.
Confirming should cause the page to be deleted while clicking cancel should do nothing.


jcheringer (approved):
LGTM!

PR #170, opened 2022-03-02
Simplify matchRoute function

jcheringer:
## Summary

This is just a small refactor to make the code a bit cleaner and easy to read.

## Test instructions

We just need to make sure the Router is working as expected by navigating on the Editor and Project Renderer routes.

### DIFF
--- a/packages/router/src/util.js
+++ b/packages/router/src/util.js
@@ -5,11 +5,10 @@ import {
 	fromPairs,
 	isNumber,
 	isUndefined,
-	keys,
 	map,
 	split,
 	trimStart,
-	zipObject,
+	mapValues,
 } from 'lodash';
 import RouteParser from 'route-parser';
 
@@ -93,10 +92,7 @@ export const matchRoute = ( pattern, path ) => {
 	const route = new RouteParser( pattern );
 	const match = route.match( path );
 
-	return zipObject(
-		keys( match ),
-		map( match, ( param ) =>
-			isNumber( param ) ? parseInt( param, 10 ) : param
-		)
+	return mapValues( match, ( value ) =>
+		isNumber( value ) ? parseInt( value, 10 ) : value
 	);
 };
--- a/packages/router/src/util.js
+++ b/packages/router/src/util.js
@@ -3,10 +3,10 @@
  */
 import {
 	fromPairs,
-	isNumber,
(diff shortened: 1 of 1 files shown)


CGastrell (approved):
Works for me :) :shipit:

ice9js (changes requested):
So actually, unfortunately this whole expression has broken the router.

`matchRoute` used to return `null` is no match is found when no match was found and the `<Route />` component depends on that behavior.

To be honest, the more I think about it, we should just revert #167 altogether.

`toNumber` works with different bases so for example `0b1` (pun intended) translates to `1`.  
The router is also used on our project renderer which uses hexadecimal project IDs which aren't converted to numbers as long as they're not prefixed with `0x` but hexadecimal values that happen to only contain digits in 0-9 range would still be converted to a decimal integer which while it might work is not entirely great either.

If anything, I guess the proper way would be to define the type on the route itself as sometimes even when the value is numeric, it's not necessarily meant to be a number. So I'd say let's remove it and just stick with strings for consistency - especially since we don't actually need this for anything?

jcheringer:
> So actually, unfortunately this whole expression has broken the router.
> 
> `matchRoute` used to return `null` is no match is found when no match was found and the `<Route />` component depends on that behavior.
> 
> To be honest, the more I think about it, we should just revert #167 altogether.
> 
> `toNumber` works with different bases so for example `0b1` (pun intended) translates to `1`. The router is also used on our project renderer which uses hexadecimal project IDs which aren't converted to numbers as long as they're not prefixed with `0x` but hexadecimal values that happen to only contain digits in 0-9 range would still be converted to a decimal integer which while it might work is not entirely great either.
> 
> If anything, I guess the proper way would be to define the type on the route itself as sometimes even when the value is numeric, it's not necessarily meant to be a number. So I'd say let's remove it and just stick with strings for consistency - especially since we don't actually need this for anything?

Well, the idea here was to make things a bit simpler but, considering the time we spent discussing this solution and yet we didn't find a reliable solution, I'd say things are not getting simpler. 😅 
It's fine to me if we decide to discard this PR and revert the #167.

ice9js:
Yeah, let's revert it  and postpone this until there actually is a need for it. Meanwhile we can continue the fun debate, lol.

PR #172, opened 2022-03-03
Move Question Wrapper default border styles

jcheringer:
## Summary

The purpose of this PR is to move MC Question and Text Question default `borderRadius` and `borderWidth` attributes to the CSS of the Question Wrapper.
This will allow us to override the default border styles for the new themes.
If the user changes the block settings, this will be added to the Block's `style` attribute, hence having higher priority. 

## Test Instructions

* Add an MC Question and a Text Question block
* The default styles for the blocks should look the same as before those changes.


ice9js (approved):
Seems to work well for me 👍

PR #173, opened 2022-03-03
Revert "Attempt to cast int on number-ish param match"

ice9js:
Reverts Automattic/crowdsignal-ui#167.

Unfortunately, that PR caused a regression in the router where it now returns `{}` instead of `null` if a route doesn't match the pattern.
Causing dependent code to 'think' it does match.

# Testing

The router should go back to working correctly - test it by trying to access project resutls at `/project/:projectId/results`.

### DIFF
--- a/packages/router/src/util.js
+++ b/packages/router/src/util.js
@@ -1,16 +1,7 @@
 /**
  * External dependencies
  */
-import {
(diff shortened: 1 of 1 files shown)


jcheringer (approved):
🚢

PR #175, opened 2022-03-04
Add rating scale question block

CGastrell:
This PR introduces the new Rating Scale question block.

The block still lacks a myriad features and style adjustments, it works in conjunction with D76521-code to handle question creation and responses on the backend.

For the scope of the new block this PR should suffice, further customizations and features to be added on upcoming PRs

## Test instructions
Apply backend patch and checkout this branch. Create/edit a project and add a Rating block


apps/project-renderer/src/components/app/index.js
```diff
@@ -32,9 +34,7 @@ const App = ( {
 	const [ content, setContent ] = useState( [] );
 	// eslint-disable-next-line
```
ice9js, line 35: Let's drop this one as well while we're at it?

packages/block-editor/src/rating-scale-answer/edit-button.js
```diff
@@ -0,0 +1,52 @@
...
+	onChange,
+	onReplace,
+	onSplit,
+	onDelete,
+} ) => {
+	return (
+		<Button attributes={ attributes } as={ 'div' } className={ className }>
+			<ButtonContent>
+				<RichText
+					placeholder={ __( 'Enter an answer', 'blocks-editor' ) }
```
ice9js, line 36: Typo: the i18n domain is `block-editor` not `blocks-editor`.

packages/block-editor/src/rating-scale-answer/edit-checkbox.js
```diff
@@ -0,0 +1,44 @@
...
+	onReplace,
+	onSplit,
+	onDelete,
+} ) => (
+	<FormCheckbox.Label
+		as={ 'div' }
+		className={ className }
+		style={ useColorStyles( attributes ) }
+	>
+		<FormCheckbox type={ multipleChoice ? 'checkbox' : 'radio' } />
```
ice9js, line 27: Hmm... I'm not sure why you'd want a multiple choice rating but feel free to enlighten me.

packages/block-editor/src/rating-scale-answer/style.scss
```diff
@@ -0,0 +1,5 @@
+.wp-block[data-type="crowdsignal-forms/rating-scale-question"]:not(.is-selected):not(.has-child-selected) {
```
ice9js, line 1: Actually, this should go into the theme compatibility layer. If it's a default to be shared by all themes, put it in `base/editor.scss`.

packages/block-editor/src/rating-scale-answer/index.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+import { MultipleChoiceAnswerIcon } from '@crowdsignal/icons';
+import attributes from './attributes';
+import EditRatingScaleAnswer from './edit';
+
+const name = 'crowdsignal-forms/rating-scale-answer';
+
+const settings = {
+	title: __( 'Answer', 'blocks' ),
+	description: __( 'Add more answer options to your question', 'blocks' ),
```
ice9js, line 17: Nit: `i18n` domain doesn't match the package name.

packages/block-editor/src/rating-scale-question/attributes.js
```diff
@@ -0,0 +1,44 @@
...
+		default: 5,
+	},
+	// Style attributes, should follow the name scheme supported by @crowdsignal/styles helpers.
+	backgroundColor: {
+		type: 'string',
+	},
+	borderColor: {
+		type: 'string',
+	},
+	borderRadius: {
```
ice9js, line 26: I believe the border radius and width props should default to `''` since #172.

packages/block-editor/src/rating-scale-question/edit.js
```diff
@@ -0,0 +1,80 @@
...
+ * Internal dependencies
+ */
+import { useClientId } from '@crowdsignal/hooks';
+
+/**
+ * Style dependencies
+ */
+import { QuestionHeader, QuestionWrapper } from '@crowdsignal/blocks';
+// import Toolbar from './toolbar';
+import Sidebar from './sidebar';
```
ice9js, line 18: I don't think these are style dependencies?

packages/blocks/src/components/question-wrapper/index.js
```diff
@@ -56,6 +56,12 @@ const Content = styled.div`
 			margin-bottom: 0;
 		}
 	}
+
+	&.is-horizontal {
```
ice9js, line 60: Technically this works... I wonder though if it's maybe better to create a separate `<Row />` component for this or something? Would make adjusting some of these like `justify-content` easier in the future.  
And this will definitely not cut it for something like a matrix setup.

packages/blocks/src/rating-scale-answer/button.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+import { RawHTML } from '@wordpress/element';
+import styled from '@emotion/styled';
+
+/**
+ * Internal dependencies
+ */
+import { Button, FormCheckbox } from '../components';
+
+const ButtonContent = styled.span`
```
ice9js, line 12: As this is basically copied over from `multiple-choice-answer`, we should think about making this into its own thing maybe.
But could be considered outside of the scope for this PR.

packages/theme-compatibility/src/base/editor.scss
```diff
@@ -53,6 +53,17 @@
 	}
 
 	.crowdsignal-forms-question-wrapper__content {
+
+		&.is-horizontal {
+			display: block; // abort flex to get full width!
```
ice9js, line 58: Or `flex: 1;` ? Also, why is this necessary here? Shouldn't the styled component styles take care of it already?

ice9js (changes requested):
Overall, I think it looks nice. I left a few small comments and I will need an explanation from someone about how a multiple choice rating makes sense?

packages/block-editor/src/rating-scale-answer/style.scss
jcheringer: I'd ask why this is necessary?

CGastrell:
Except for the style and row comments, all the other *quirky* things (i18n domain, wrong import section, etc) come from the copy/paste from MC Question and Answer blocks. I'll try to address those on both instances.

As for the styles: I think some `Row` thing makes sense. Not sure if that would only be for a matrix or it's already justified here. It's also true that when I faced the idea of this to *pave the road* for the matrix, a lot of things would need to be changed, specially some row/table behaviors that I still need to figure out (and don't have anything to test with). Most challenging one being how to make *cells* of the `Row` to share the width their above/below  neighbors (this was also a thing on the legacy editor, but we ended up using a full table there).

packages/block-editor/src/rating-scale-answer/edit-checkbox.js
CGastrell: Matrixes do have a "multiple answer" option. While this is not the case for ratings, I'd thought we could keep that here if we are to use this component in the future.

CGastrell: Then again, the file is not used by any component, so I might just remove the whole thing

jcheringer (changes requested):
@CGastrell One little thing I notice is that when you add the Rating block on a blank project, it doesn't include the Submit button.
We need to map the new block on `apps/dashboard/src/components/editor/auto-submit-button.js`.
Other than that, looks good to me.
About the `is-horizontal` thing, if we really want to handle this now, I'd do it on a separate PR.

digitalwaveride:
Nice nice nice Christian!!!

The only thing ....  that emoji styling would be a very nice touch to have... 

image

jcheringer (approved):
LGTM!

PR #176, opened 2022-03-07
Set project renderer's bottom padding to 80px

ice9js:
This patch adjusts the project's renderer default bottom padding to 80px.

Solves c/bvbpKxxD-tr.

# Testing

Verify the padding is updated.


jcheringer (approved):
Works for me! 🚢

PR #177, opened 2022-03-07
Update Editor to prevent side effects from new Themes

jcheringer:
## Summary

This PR updates a few Editor CSS styles to increase their specificity and prevent side effects from new themes implementation.

Bonus: Removes the default input height and allows it to be empty. 
The default min-height is defined on the block's style. Removing the default attribute value will allow the themes to override the styles when necessary.

## Test Instructions

* On the Editor, take a look at the styles of the Header, Right and Left sidebars
* The layout should look unaffected
* Add a Text Input form and change its input height on the block's settings
* Save the project and reload the page
* Go to the Text Input settings again and clear the input height property
* You should be able to save the project and the block should now use the default input height defined on the styles

### DIFF
--- a/apps/dashboard/src/components/editor/styles/editor.js
+++ b/apps/dashboard/src/components/editor/styles/editor.js
@@ -135,6 +135,10 @@ export const EditorWrapper = styled.div`
 				}
 			}
 		}
+
+		a {
+			color: var( --color-primary-50 );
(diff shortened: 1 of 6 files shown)


digitalwaveride (approved):
Works as advertised!
🚀 

Thanks João!

PR #180, opened 2022-03-10
Switch sidebar tabs depending on if any blocks are currently selected

ice9js:
This patch adds a feature where the sidebar will toggle between `Document` and `Block` tabs depending on if any blocks are currently selected.

Solves c/1jlmCWMA-tr.

# Testing

Go to the editor and verify the sidebar works as described.


digitalwaveride (approved):
A delight!
Makes things so much better 😁 !
Thanks Kuba!

PR #181, opened 2022-03-11
Add/email form block

TheSteveK:
This PR adds a block variation to the text form block.

A new file was created that holds an array of variation objects.  Adding additional variations to the text input form can be done by adding new objects to the text form block.

Another new file, validation.js was created to conveniently hold variation code and corresponds with the validation attribute in the variations file.

A function call was added to the block/text-input/index.js file that will run the function named in the variations object to run any sort of form field validation related to the new type of field.

This last commit fixes the last of the suggestions in the previous PR, and this is ready for another look.

Testing instructions:

Checkout branch and yarn install/start
Add a new project via https://crowdsignal.localhost:9000/project
Add the new email block via the block library.
Feel free to add any other blocks just for fun.
Preview the project.  Enter something that's a mailformed address (the validator expects an @ and a valid . domain name) and click away from the email field.  A red error message will appear and the form will not submit.
Update the field with a vaild email address, see the error go away and submit will now work.


TheSteveK:
adding this as a note to self... I need to see if it's also possible to change the icon in the toolbar for variations: 
Screen Shot 2022-03-11 at 10 12 33 AM

packages/block-editor/src/text-input/attributes.js
```diff
@@ -29,4 +29,7 @@ export default {
 		type: 'string',
 		default: '100%',
 	},
+	validation: {
+		default: null,
```
ice9js, line 33: Don't we need a `type` to be defined here as well?

packages/block-editor/src/text-input/index.js
```diff
@@ -9,6 +9,7 @@ import { __ } from '@wordpress/i18n';
 import { TextInputIcon } from '@crowdsignal/icons';
 import attributes from './attributes';
 import Edit from './edit';
+import Variations from './variations';
```
ice9js, line 12: Capitalized variables are usually reserved for and suggest components and/or classes but we're just dealing with a regular object here. I'd suggest we just rename it to `variations`.

packages/block-editor/src/text-input/variations.js
```diff
@@ -0,0 +1,27 @@
+import { __ } from '@wordpress/i18n';
```
ice9js, line 1: Missing external/internal dependencies docblocks.

packages/block-editor/src/text-input/variations.js
```diff
@@ -0,0 +1,27 @@
...
+
+import { EmailInputIcon } from '@crowdsignal/icons';
+
+export default [
+	{
+		name: 'email',
+		isDefault: false,
+		title: __( 'Email Input' ),
```
ice9js, line 9: Missing i18n domain argument here and on the next line.

packages/block-editor/src/text-input/variations.js
```diff
@@ -0,0 +1,27 @@
...
+		description: __( 'A form field for email addresses' ),
+		icon: <EmailInputIcon />,
+		attributes: {
+			label: 'Enter email',
```
ice9js, line 13: Missing i18n.

packages/block-editor/src/text-input/variations.js
```diff
@@ -0,0 +1,27 @@
...
+			validation: 'emailValidation',
+			placeholder: 'name@yourgroovydomain.com',
+		},
+
+		keywords: [
+			__( 'email input', 'block-edtor' ),
```
ice9js, line 19: Typo in the i18n domain name here and the lines below.

packages/blocks/src/text-input/index.js
```diff
@@ -12,14 +12,20 @@ import { isEmpty } from 'lodash';
 import { useColorStyles } from '@crowdsignal/styles';
 import { ErrorMessage, FormInputWrapper, FormTextInput } from '../components';
 import { useField } from '@crowdsignal/form';
+import validator from './validations';
 
 const TextInput = ( { attributes, className } ) => {
+	//console.log(validator[attributes.validation]());
```
ice9js, line 18: Let's remove any logs/comments like this before we merge it.

packages/icons/src/email-input.js
```diff
@@ -0,0 +1,16 @@
...
+	<svg
+		width="24"
+		height="24"
+		viewBox="0 0 24 24"
+		fill="none"
+		xmlns="http://www.w3.org/2000/svg"
+	>
+		<path
+			d="M18 6.75H6C5.30964 6.75 4.75 7.30964 4.75 8V16C4.75 16.6904 5.30964 17.25 6 17.25H18C18.6904 17.25 19.25 16.6904 19.25 16V8C19.25 7.30964 18.6904 6.75 18 6.75Z"
+			stroke="#1E1E1E"
```
ice9js, line 11: We generally prefer to style these using CSS so the stroke attribute can be omitted here and in the other `<path />` below.

ice9js (changes requested):
I left some minor comments but overall this is looking good!
I like how you separated the validation code - for now it can remain where it is, eventually we might want to bump it up or put in its own package so some of these can potentially be reused by other blocks. But let's not worry about that now.

There is one caveat - what if a question would have multiple validators that can potentially be toggled on or off?
While we don't need to implement this right here right now, it'd be good to make sure the format of `attributes.validation` allows for it later without affecting existing blocks.

packages/block-editor/src/text-input/index.js
TheSteveK: calling it "variations" was messing with it loading the attribute properly.  I went ahead to change the import to variation to keep in line with the conventions

digitalwaveride:
Works! 🚀 

Thanks so much Steve!!!

One last little thing, the name of the block: Let's call the block "`Email Form`".

So all these types of blocks end with "Form"

We would have later... 

Text Input Form
Email Form
Url Form
... 

The other type of blocks are "Questions" which are these bigger "card type" layouts.

CGastrell:
> Works! 🚀
> 
> Thanks so much Steve!!!
> 
> One last little thing, the name of the block: Let's call the block "`Email Form`".
> 
> So all these types of blocks end with "Form"
> 
> We would have later...
> 
> Text Input Form Email Form Url Form ...
> 
> The other type of blocks are "Questions" which are these bigger "card type" layouts.

I like the idea of having a sort of standardized naming. I'd add "input" also to our suffix. One extra word, one easier way to find the block you're looking for. And seems more specific at what it does as well, "email form" can easily be taken as an "action" rather than an input. What do you think @digitalwaveride ?

Text Input Form
Email Input Form
URL Input Form

TheSteveK:
Ok, there's actually two fields, a name and a title.

I updated the name to match the format of the parent text input block to be:
crowdsignal-forms/email-input

And the title is now following Christian's suggestion:
Email Input Form

packages/block-editor/src/text-input/index.js
```diff
@@ -27,6 +28,7 @@ const settings = {
 	],
 	icon: <TextInputIcon />,
 	edit: Edit,
+	variations: variation,
```
ice9js, line 31: Nit: you could also import the variable as `variations` ad just put `variations,` in here making it slightly shorter.

packages/blocks/src/text-input/index.js
```diff
@@ -20,6 +21,11 @@ const TextInput = ( { attributes, className } ) => {
 			if ( attributes.mandatory && isEmpty( value ) ) {
 				return __( 'This field is required', 'blocks' );
 			}
+			if ( attributes.validation && ! isEmpty( value ) ) {
+				for ( let i = 0; i < attributes.validation.length; i++ ) {
```
ice9js, line 25: Nit: Nowadays JS also offers `for ... of ...`, which would be a little shorter and assigns the array elements directly.

ice9js (changes requested):
Looks good - just the return statement inside the validation loop is kind of wrong so would be nice to resolve that one. (Although it doesn't affect anything if you just have a single validator).

ice9js (approved):
Looks great, let's get this merged :)

PR #182, opened 2022-03-11
Refactor editor settings to not use iso.blocks.disallowBlocks property

ice9js:
This fix replaces `iso.blocks.disallowBlocks` with `iso.block.allowBlocks` in `editor/editor.js` to hide our question blocks on the confirmation page.  
It seems that using `disallowBlocks` at all causes that to take precedence over `allowBlocks` and unwanted blocks appear again. With this fix applied, that should no longer be the case.

# Testing

Apply the patch and verify only the blocks on our allow list are displayed in the editor inserter.
No Crowdsignal question or submit button blocks should be available when on the confirmation page.


digitalwaveride (approved):
🙏 Thanks a ton Kuba!

Works 🚀 !!!

PR #183, opened 2022-03-11
Fix cover block preview

ice9js:
This patch fixes page and template previews for full size cover blocks.  
The cover block uses `min-height: 100vh` property to resize itself which works in the browser but was causing issues in the previews where `vh` is based on the current iframe size which is based on the current iframe content - which leads to a race condition and infinite resizing of the iframe.

By putting a cap on the preview height we can prevent it. I choose the same number as used internally by `<BlockPreview />` - unfortunately their fix doesn't quite work in this case.

Note: Because of this, the preview doesn't quite match the editor content in this case but it should be close enough for what it needs to be.

# Testing

- Add a cover block and upload an image. (Or start by adding an image block and converting it to a cover block).
- Select 'Toggle Full Height' from the toolbar and observe the page preview.
- The block should resize just once and stay there compared to continuous enlargement that happened previously.

### DIFF
--- a/apps/dashboard/src/components/editor/styles/editor.js
+++ b/apps/dashboard/src/components/editor/styles/editor.js
@@ -13,6 +13,13 @@ export const editorGlobalStyles = css`
 	html.interface-interface-skeleton__html-container {
 		width: 100% !important;
 	}
+
+	html.block-editor-block-preview__content-iframe
+		.is-root-container.block-editor-block-list__layout {
+		// Value chosen to match
+		// https://github.com/WordPress/gutenberg/blob/200a5d700e19d9f5978877cd299d582b1480a8d2/packages/block-editor/src/components/block-preview/auto.js#L20
+		max-height: 2000px;
+	}
 `;
 
 export const EditorLayout = styled.div`


digitalwaveride (changes requested):
Nice Kuba!
"Endless zooming" in the previews column is fixed. 👍 
It has issues though fro me with displaying the "full width" characteristic in the previews... 
Feels like the preview supports just "Align Center" (all pages below have cover block Full Width -> .../project/2743672)

image

CGastrell (approved):
The way I see it, we need to compromise as we can't make a full preview of a too tall page. This should do by now IMHO

digitalwaveride:
> The way I see it, we need to compromise as we can't make a full preview of a too tall page.

The "windows" in the paging column, should not symbolize a "full-page preview", otherwise these windows can't stay the same height, and the preview windows would be ridiculously long.

The windows should preview a viewport of a "laptop device".
We could show simply the viewport of the top of the page ... the state without scrolling. Everthing that is "below the fold" is not visible in that preview.

ice9js:
Going to merge it to not hold up what's already fixed any longer.  
As for the full width thing... we need to rebuild the preview component at some point anyway which would be a good opportunity to see if we can address that. As of right now, I don't think 'full width' previews are supported.

PR #185, opened 2022-03-15
Create a Twenty Twenty Two dark theme

jcheringer:
## Summary

We want a `Twenty Twenty Two` dark version.

Depends on: D76911-code

## Test Instructions
* Add the theme parameter to the URL: `?theme=twentytwentytwo-dark`
* You should see the theme using a dark background and white text color

## Screenshots

image

### DIFF
--- a/packages/theme-compatibility/src/base/editor.scss
+++ b/packages/theme-compatibility/src/base/editor.scss
@@ -45,6 +45,10 @@
 			margin-top: 0;
 			margin-bottom: 0;
 		}
+
+		&:not(:first-child) {
+			margin-left: var(--wp--style--block-gap, 2em);
+		}
 	}
 
 	.crowdsignal-forms-question-wrapper {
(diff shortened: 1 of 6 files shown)


digitalwaveride (changes requested):
Thanks so much João!!!

Just one little thing, in the text question block, the outline of the input fields with a border of 2px feels like too much. Should we reduce it to 1px? 

Before:

image


After:

image



And I was just wondering if we should keep the "light theme" as well... having a TwentyTwentyTwo light and dark theme?

jcheringer:
Hey @digitalwaveride!
I did some changes to keep the current `TwentyTwentyTwo` as is and add a dark version of the theme.
Also did some changes to match the suggested style.
Note that, since we needed some backend changes, to test it properly, you'll need the changes on D76911-code.

digitalwaveride (approved):
😍 Perfect!!!

PR #188, opened 2022-03-17
Update MC Answer button to use outline style

jcheringer:
## Summary

Currently, all buttons are treated equally. 

The button styling of the answer options needs to differ from the Submit button. 
 Otherwise, the main CTA is not obvious enough.  

Ref: c/vRbB6nCJ-tr

## Test Instructions

* Open or create a project and add an MC Question block
* You should see the answers buttons using an outline style

## Screenshots

image

### DIFF
--- a/packages/block-editor/src/multiple-choice-answer/edit-button.js
+++ b/packages/block-editor/src/multiple-choice-answer/edit-button.js
@@ -40,6 +40,7 @@ const EditButtonAnswer = ( {
 			style={ {
 				width,
 			} }
+			outline
 		>
 			<ButtonContent>
 				{ multipleChoice && <FormCheckbox type="checkbox" /> }
--- a/packages/blocks/src/components/button/index.js
+++ b/packages/blocks/src/components/button/index.js
@@ -40,14 +40,18 @@ const Button = ( {
 	attributes,
 	children,
 	className,
+	outline,
 	style = {},
 	...props
 } ) => (
 	<StyledButtonWrapper
 		className={ classnames(
-			'crowdsignal-forms-button',
 			className,
-			'wp-block-button'
+			'crowdsignal-forms-button',
+			'wp-block-button',
+			{
+				'is-style-outline': outline,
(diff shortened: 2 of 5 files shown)


digitalwaveride:
Nice João! 🙏 

Looks good! One little optimization for the "select state":

How about giving it a background like this:

image


And one other thing I noticed, not sure if related to this PR at all, but when I select the "multiple choice" option, I see it in the editor, but don't see it in the Preview or Published state. There it shows as a "single choice" question:

image

image

jcheringer:
Hey @digitalwaveride!
I think changing the style for the selected answer makes sense but, since we're dealing with multiple themes, changing the answer background could be a bit tricky.
I pushed a change that will make the selected answer use the default button style, this way we can take advantage of what already exists on the theme.
Please, take a look and let me know what you think.

About the multiple-choice issue, it worked fine for me. 😬

digitalwaveride:
Thanks João!

Leven
TwentyTwentyTwo
Blockbase

... work perfectly! 


The TwentyNineteen is unfortunately acting up for me... 

image


The Quadrat theme has a weird effect after click, as on hover the button is not filled yet, then on cursor out the button flickers to the "active state". Which feels off... 

image

image

jcheringer:
Thomas, thanks for testing it again!
My last commit should address the TwentyNineteen issues.
About the Quadrat theme, following our external conversation, we need to decide if this is something we want to do as part of our compatibility changes.
Anyway, even if we decide to do those changes, can we consider them outside the scope of this PR?

CGastrell (approved):
Worked nicely on my end :shipit:

digitalwaveride (approved):
>  Anyway, even if we decide to do those changes, can we consider them outside the scope of this PR?

That's fine! 
Thanks for fixing 2019 🙏 !!!

Let's address Quadrat theme behavior (or as discussed maybe the behavior for all themes for these "selected", and "selected on hover" states) in a separate PR

PR #189, opened 2022-03-17
Handle form array values as multiple entries of the same name

CGastrell:
This PR changes the submit function on the project-renderer app. It does so to consider array values as multiple entries of the same name, the way PHP usually receives those:
formarrays

Of course, we can leave this alone here and handle things on backend, but AFAIR, form POST requests have always worked this way.

Let me know what you think @ice9js @jcheringer 

## Test instructions
Create a project with a MC question. Make it *checkbox* so multiple options can be selected. Visit the project and submit a response, check the results page to confirm the selections were properly recorded


ice9js (approved):
LGTM!

PR #190, opened 2022-03-17
Fix themes on page previews

ice9js:
This is a partial fix for themes not updating on page previews. This patch will force the page navigation column to re-render if the current theme changes.  
This fixes the issue for themes but unfortunately other issues where certain block styles might not load correctly inside the preview. It's also not an ideal experience as it takes a few seconds for all previews to re-render from scratch so it's definitely noticeable.

The core issue is related to Gutenberg itself not allowing the preview styles to be updated once the preview is rendered ([link](https://github.com/WordPress/gutenberg/blob/f2161e246b9fdd9a2a56e7552b0b28050f1a5302/packages/block-editor/src/components/iframe/index.js#L230)).  
A proper fix would involve creating our own Preview component which does allow for this which would allow us to address remaining edge cases as well as provide a much smoother experience.

# Testing

Try changing themes using the theme modal in the editor. Page previews should be updated with the theme styles.

### DIFF
--- a/apps/dashboard/src/components/editor/editor.js
+++ b/apps/dashboard/src/components/editor/editor.js
@@ -111,7 +111,7 @@ const Editor = ( { project } ) => {
 				/>
 			) }
 
-			<PageNavigation />
+			<PageNavigation key={ `nav_${ editorTheme }` } />
 			<EditorWrapper
 				as={ IsolatedBlockEditor }
 				key={ editorId }
--- a/apps/dashboard/src/components/editor/styles-resolver.js
+++ b/apps/dashboard/src/components/editor/styles-resolver.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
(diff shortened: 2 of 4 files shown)


digitalwaveride (approved):
Nice Kuba! 😻 

Yes, it's not perfect, but for now... a great fix for allowing us to move on!

💪 🙏

ice9js:
I added a small fix to make sure the page previews background matches the theme and updated leven to use `--wp-custom-color-background` variable that seems to be the standard for newer themes, as discussed with @jcheringer.

jcheringer (approved):
LGTM! :shipit:

digitalwaveride:
Just found this issue for "Starting from Scratch" template:

image

ice9js:
>Just found this issue on "Start from Scratch" template...

Already been addressed :)

PR #191, opened 2022-03-18
Adjust rating scale styles and meta

CGastrell:
This PR is still a WIP

* Move is-horizontal out of the styled component (QuestionWrapper.Content) and into the base scss
* Add icon for the rating scale question and rating scale answers. While the latter exists, it can't be used/added/removed, so not sure how useful that icon is there
* Use new outline button styles (props to @jcheringer )
* Show the style options on the correct order: Text Buttons - Emojis

Pending tasks (still unsure how to roll these out):

- [ ] Show a single (big) sample on the style selection of the block (right sidebar)
- [ ] When style changes, replace values of the rating scale with some emoji map (1 through 5?) and *vice versa*
- [x] Wrap the button to force a transparent border on emoji style
- [x] isolate `:last-child` selector affecting the horizontal layout


CGastrell:
Abandoned in favor of #193

PR #192, opened 2022-03-18
Update Templates modal to allow theme selection

jcheringer:
## Summary

The purpose of this PR is to allow the user to select their project theme from the Templates modal.

Ref: c/bjJkPbiT-tr

Depends on: D77208-code

## Test Instructions

* Start a new project
* You should be able to change the project theme from the Templates modal

## Screenshots

image

### DIFF
--- a/apps/dashboard/src/components/editor/editor.js
+++ b/apps/dashboard/src/components/editor/editor.js
@@ -100,7 +100,10 @@ const Editor = ( { project } ) => {
 			/>
 
 			{ showWizard && (
-				<NewProjectWizard onSelect={ handleSelectTemplate } />
+				<NewProjectWizard
+					onSelect={ handleSelectTemplate }
+					onChangeThemeClick={ handleOpenThemesModal }
+				/>
 			) }
 
 			{ showThemesModal && (
--- a/apps/dashboard/src/components/editor/editor.js
+++ b/apps/dashboard/src/components/editor/editor.js
@@ -69,7 +69,8 @@ const Editor = ( { project } ) => {
 	};
 
 	const handleSelectTheme = ( theme ) => {
-		setProjectTheme( theme );
+		const autoSave = ! showWizard;
+		setProjectTheme( theme, autoSave );
 		setShowThemesModal( false );
 	};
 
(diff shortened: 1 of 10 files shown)


CGastrell (changes requested):
Not working for me. But, from the "test instructions" I can't be sure I'm on the right scope, so, here's what happened:

- Patched
- Create new project
- Chose a theme, got to editor with theme applied
- Added some changes, click preview: preview shows Leven theme
- Published (an unrelated regression bug occurred) and previewed again, still Leven theme
- Checked meta on DB: both themes show Leven

jcheringer:
> Super nice João! 🙏
> 
> Two little things that might not be related to this PR:
> 
> #### Remembering the chosen theme
> When you create the next new project, would it be possible to remember the last choice you made?
> 
> That means, obviously we don't want to "overwrite" or change themes of existing projects, just for the new project we want to use by default the theme based on the last theme choice you made. Assuming you have your favourite theme, selecting it over and over again would be annoying.

That should be possible. We just need to discuss the best solution because, currently, when you start a new project, we don't make any requests to the backend.


> #### The Preview of the templates are not adjusting to theme
> Would that be already doable by rendering? Or do we need to add screenshots for each of the templates fitting to the themes?

That should be fixed by #190

jcheringer:
> Not working for me. But, from the "test instructions" I can't be sure I'm on the right scope, so, here's what happened:
> 
> * Patched
> * Create new project
> * Chose a theme, got to editor with theme applied
> * Added some changes, click preview: preview shows Leven theme
> * Published (an unrelated regression bug occurred) and previewed again, still Leven theme
> * Checked meta on DB: both themes show Leven

Hey @CGastrell, my last commit should solve the saving/publishing inconsistencies.
Note that, now you'll also need code from D77208-code

digitalwaveride:
Regarding "Remembering the chosen theme":

Created a new Trello Card for this, and prioritized it for the next release.

I am good with this, whenever @CGastrell gives a thumbs up.

Thanks @jcheringer !!!

CGastrell (approved):
Works nicely!

PR #193, opened 2022-03-18
Fine tuning the Rating Scale block

CGastrell:
This PR is a non-hostile take over #191 with fine tuning on the Rating Scale block

It includes simple meta stuff, like icon, description and keywords for the Rating Scale block as well a styles and functionalities particular to the block. A backend change has been deployed to deal with the block's requirements.

* Move is-horizontal out of the styled component (QuestionWrapper.Content) and into the base scss
* Add icon for the rating scale question and rating scale answers. While the latter exists, it can't be used/added/removed, so not sure how useful that icon is there
* Use new outline button styles (props to @jcheringer )
* Show the style options on the correct order: Text Buttons - Emojis

Pending tasks have been mostly addressed, yet one remains that I think we can tackle on its own:

- [ ] Show a single (big) sample on the style selection of the block (right sidebar)
- [x] When style changes, replace values of the rating scale with some emoji map (1 through 5?) and *vice versa*
- [x] Wrap the button to force a transparent border on emoji style
- [x] isolate `:last-child` selector affecting the horizontal layout

## Test instructions
Checkout and create a new project. Insert the new Rating Scale block. You can switch between the block styles "Text" and "Emoji". Most noticeable difference is emojis default to a borderless button and a larger font-size (so emojis don't look microscopic). Each style will retain the user set value.


digitalwaveride:
Super nice Christian!!!! Yay! We are close!

Some little styling finetuning things:

On the **Leven theme**:
The "text button select state" ... the buttons are jumping a bit as they don't have the same 

image


**Twenty nineteen** 
is showing some outlines for the emojis...
image

**Blockbase  & TwentyTwentyTwo's & Quadrat theme**
The selected state background is too big ... it should have about the same size of the text button. The increased font-size for the emoji increased the whole "button", but to keep it consistent visually, the "button space" around the emoji should keep the same size.

image


image

CGastrell:
Yeah, the "width" change on Leven and 2019 is something related to the "outline" button style. I might have to work on it at a custom level. Right now, the buttons simply use `is-outline` and I thought that would be enough.

I guess the same goes for the observations on the other themes.

Then again, these are all sizes and spacing established by the theme and components, if we start customizing every single bit of it (widths, borders, paddings) then it all starts to *not make sense* as they fall into a "button but not button" or "normalized space but not normalized" kind of thing.

All that said, if these are the only concerns, then I'd merge it and address these on a separate PR

digitalwaveride:
> All that said, if these are the only concerns, then I'd merge it and address these on a separate PR
Sure! Feel free to merge and let's address that in a separate PR.

PR #195, opened 2022-03-21
Allow page navigation to scroll when it extends past the window

ice9js:
This patch addresses some of the issues related to the editor's page navigation column:

- When the previews exceed window height, the column should become scrollable.
- Fixes an issue when certain elements of the column resize or collapse vertically.
- Only the previews part of the column is scrollable, the header stays in place.
- Confirmation page now features a regular page number instead of `CP`.

Solves c/5ZNsBrIL-tr and c/PxdeY3RU-tr.

# Testing

Open the editor and verify the page navigation column works as described on smaller screen sizes.

### DIFF
--- a/apps/dashboard/src/components/editor/page-navigation.js
+++ b/apps/dashboard/src/components/editor/page-navigation.js
@@ -21,6 +21,7 @@ import { STORE_NAME } from '../../data';
  */
 import {
 	PageNavigationAddButton,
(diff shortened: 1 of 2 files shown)


ice9js:
I couldn't quite reproduce it so merging it as it's just a styling issue but will keep an eye out on it.

PR #196, opened 2022-03-21
Open the inserter when creating a new project

ice9js:
This patch will force the inserter panel open when the editor is first opened for a new project.

Solves c/J5XKzxfN-tr.

# Testing

Start a new project - the inserter should be open by default.
When opening an existing project, the inserter should remain closed.


CGastrell (approved):
Works nicely, except on the first auto-save.

- new project, inserter shows
- click on one of our blocks, block and submit are inserted, no auto-save is triggered (?)
- close inserter, make a change, wait for auto-save
- **once auto-save is done, inserter opens again**

Works for me, I'll leave it up to you if you want to address that misbehave here or on a different PR

PR #200, opened 2022-03-24
Add an editor error notice

ice9js:
This patch implements an editor error notice to properly inform users about any issues with the editor itself.  
Specifically, at the moment, it should let the user know if there's been an issue with the projects' save requests and urge them to check the connection.

This also fixes a bug where project's `saveAndUpdateProject` action generator was preventing the error handling part of the editor from ever getting triggered - so it'd actually even indicated that your changes have been saved when they weren't.

Solves c/95Xou7RS-tr.

# Testing

- Start a new project or open an existing one.
- Turn your browser into offline mode.
- Make some changes.
- The save request should fail and you should see the error notice at the top of the editor.
- The editor toolbar should still reflect the unsaved state (save draft/update buttons are available).
- Without making any additional changes, turn the browser connection back on and hit save.
- The request should go through and the banner should disappear.
- Verify your changes have actually been saved correctly.


CGastrell (approved):
Works as advertised!

PR #201, opened 2022-03-25
Reset useClientID registry when restoring drafts

ice9js:
This is an attempt to fix the issues with blocks' `attributes.clientId` being unnecessarily overwritten when switching to drafts for published projects.

The issue boils down to `block.clientId` being unreliable - when the content is loaded, the existing values are replaced with generated ones. It seems the same process takes place when you restore a draft (but interestingly not when you switch pages even though it's technically the same 🤷 ) so resetting the registry allows for `block.clientId` to change in those instances while retaining the same `attributes.clientId`.

Now, this does kind of work but like I said, it's a little inconsistent when it comes to switching pages and since I can't figure out why, we have to assume at some point it could fail.  
The permanent way to deal with this would be to limit `useClientId` to the scope of a single page - allowing attributes.clientId` to be duplicated across multiple pages which would also make `regenerateClientIds` in the editor state obsolete. The catch is this would require updates to the backend to deal with it - including managing existing records.

# Testing

- Create a project, add some questions, hit publish.
- Reload the page, `attributes.clientId` for the questions should stay the same.
- Make some more changes and hit save draft.
- Reload the page, `attributes.clientId` for the question should stay the same.
- Hit restore draft, `attributes.clientId` for the questions should stay the same.
- Switch between pages, `attributes.clientId` for the questions should stay the same.
- Try duplicating or copy&pasting a question from one page to another. The new copy should get a new `attributes.clientId`.

### DIFF
--- a/apps/dashboard/src/components/editor/use-editor-content.js
+++ b/apps/dashboard/src/components/editor/use-editor-content.js
@@ -7,6 +7,7 @@ import { useDispatch, useSelect } from '@wordpress/data';
 /**
  * Internal dependencies
  */
+import { useClientId } from '@crowdsignal/hooks';
 import { isPublic } from '@crowdsignal/project';
 import { STORE_NAME } from '../../data';
 import { NOTICE_UNPUBLISHED } from './notice';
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
Worked fine on the tests.

Regarding your concerns on the block.clientId thing, we might implement some of the attempts I made on https://github.com/Automattic/crowdsignal-ui/pull/186/files 

While now obsolete, the changes there add a simple clientId regenerator. Not a big thing, but making sure the block's clientId prop was forcibly different. We could put something like that on page duplicates if that's the concern.

PR #204, opened 2022-03-25
Add/url input variation

TheSteveK:
Building off the email block, this PR adds another variation of the text input form for a URL input.

To add a new block, add a variation to /block-editor/text-input/variations.js  Add a validator to blocks/text-input/validations.js (if required) and then add an icon to /icons

You then end up with something like this:

Screen Shot 2022-03-25 at 10 54 37 AM

Testing instructions:
Apply this PR and do the yarn thing then visit https://crowdsignal.localhost:9000/project and create a new project.

Click the block inserter and you should see the new URL form input block, add it and then preview your form.

Try adding an invalid URL and see if the error message appears

Correct the URL, verify the error goes away

Note, right now the validator doesn't look for http(s):// but if anyone has a strong feeling about that, the regex can be updated to require it.


digitalwaveride (commented):
Nice Steve! Works for me! 👍 

If the code review is fine ... 🚀

packages/block-editor/src/text-input/variations.js
```diff
@@ -31,4 +31,27 @@ export default [
...
+		name: 'crowdsignal-forms/url-input',
+		isDefault: false,
+		title: __( 'URL Input Form', 'block-editor' ),
+		description: __(
+			'An input field for collecting and validating website addresses.',
+			'block-editor'
+		),
+		icon: <URLInputIcon />,
+		attributes: {
+			label: __( 'Enter url', 'block-editor' ),
```
CGastrell, line 45: Should we use "URL" instead of "url"?

packages/blocks/src/text-input/validations.js
```diff
@@ -8,4 +8,14 @@ export default {
...
 		}
 	},
+
+	urlValidation: ( value ) => {
+		if (
+			! value.match(
+				/[(http(s)?):\/\/(www\.)?a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/gi
+			)
+		) {
+			return 'Please enter a valid url.';
```
CGastrell, line 18: Needs i18n.

I'd also repeat the question regarding the use of all caps on acronyms, should it be "URL"?

packages/blocks/src/text-input/validations.js
```diff
@@ -8,4 +8,14 @@ export default {
 			return 'Please enter a valid email address.';
```
CGastrell, line 8: Not part of this PR, but I just noticed this one doesn't have a translation either

CGastrell (changes requested):
Minor translation and question, but it looks good. Nice job!

TheSteveK:
I addressed @CGastrell 's notes (including URL in text).  Technically it should've been another PR, but I went ahead and added i18n to the email error message as well as the URL one.

packages/blocks/src/text-input/validations.js
```diff
@@ -1,11 +1,26 @@
...
+
 export default {
 	emailValidation: ( value ) => {
 		if (
 			! value.match(
 				/^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
 			)
 		) {
-			return 'Please enter a valid email address.';
+			return __( 'Please enter a valid email address.' );
```
CGastrell, line 13: Translation calls should also carry a domain (second argument). I think we're matching domains to the package names, so `blocks` should the domain here.

packages/blocks/src/text-input/validations.js
```diff
@@ -1,11 +1,26 @@
...
+		}
+	},
+
+	urlValidation: ( value ) => {
+		if (
+			! value.match(
+				/[(http(s)?):\/\/(www\.)?a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/gi
+			)
+		) {
+			return __( 'Please enter a valid URL.' );
```
CGastrell, line 23: Same as above, translation call lacks domain (sorry I didn't specify this on my last review)

CGastrell (changes requested):
Works nicely, I just pointed out the lack of domain parameter on the translation calls.

PR #206, opened 2022-03-25
Themes clean up

jcheringer:
## Summary

This PR is just a small clean-up on styles for the block's answers.
Instead of creating specific styles for each theme, we're creating a generic style on the base layer that relies on some CSS variables.
Then, on the themes, where necessary we can just override the values of the variables to adjust the style.

## Test Instructions
* Open or create a project
* Add at least one MC Question block and on Rating Question block
* Play with the blocks and take a look at the styles
* They should look the same as before.


CGastrell (approved):
LGTM!

PR #207, opened 2022-03-31
update/Get rid of more menu

TheSteveK:
We had the "more" settings menu active (the 3 dot menu) but had nothing to go into it since we don't support the actions in the native menu.

Declaring a false value for that menu in the settings.js file hides the menu on the editor.

Testing instructions:

Apply and go to the project editor.  Confirm there's no 3 dot menu to the right of the settings gear icon.


CGastrell (approved):
Nice! At least until we need it for something :shipit:

PR #209, opened 2022-04-01
Show confirmation when overriding public content

jcheringer:
## Summary

This PR includes the changes to show a confirmation popup whenever the user is publishing or updating content that will override the `publicContent` a might cause data loss.

**Important**: It also includes a suggested UI change to make it easier to track when we should show the confirmation popup.
WIth this change, the Visibility toggle button will only change the project visibility and won't publish project's draft content when switching from private to public.

Depends on: D78067-code

Ref: c/fy6hLLMM-tr

## Test Instructions
* Open or create a project
* Play with the Publish/Update button and the Visibility toggle
* You should see a confirmation popup whenever the content you're publishing is overriding the content already published.

### DIFF
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -29,9 +29,7 @@ const DocumentSettings = ( { project, onChangeThemeClick } ) => {
 	const { openGeneralSidebar, setIsInserterOpened } = useDispatch(
 		'isolated/editor'
 	);
-	const { saveAndUpdateProject, saveEditorChanges } = useDispatch(
-		STORE_NAME
-	);
+	const { saveAndUpdateProject } = useDispatch( STORE_NAME );
 
 	const [
 		canPublish,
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/components/editor/document-settings.js
```diff
@@ -53,7 +51,7 @@ const DocumentSettings = ( { project, onChangeThemeClick } ) => {
 			return;
 		}
 
-		saveEditorChanges( { public: true } );
+		saveAndUpdateProject( project.id, { public: true } );
```
ice9js, line 54: Having looked at D78067-code, it seems to me like this would re-publish the last-published version.  
So if you've made your project private, made some changes, and hit 'public' in the visibility menu again, it'd publish the old version.

It does kind of make sense logically... that said, you have no reference for what is actually being published when you perform the action which is kind of bad. Hence I have to ask if that's intentional or if I'm missing anything?

jcheringer: Yes, this was intentional. The reason I'm suggesting this change is to avoid putting the user in a 'no turning back' situation. Which will happen silently on our live version.
I'm also using the `publicContent` contents to validate if the project was published at some point and then show the notification when the user hits Publish/Update.

As an alternative, we could keep this logic as in the current live version but show another notification (the message should be different) warning the user that they wouldn't be able to restore their public version when changing from public to private.
But, to use this approach, we'd also need to create some flag to let us know that the project was published at some point, thus, allowing us to show the warning notification when necessary.

CGastrell (approved):
It does work as described

PR #215, opened 2022-04-12
Fix Embed block alignment on Project Renderer

jcheringer:
## Summary

This PR fixes the Embed blocks alignments, especially on the Project Renderer.

Ref: c/pD0jsWqy-tr

## Testing

* Open or create a project and add some Embed blocks
* Play with the block's alignments
* They should work properly

Note that left and right alignments use `float` (as WP) so content may overlap.


TheSteveK:
LGTM!

TheSteveK (approved):
It works! :)

PR #217, opened 2022-04-13
Update/Close Share Modal When Clicked Outside

TheSteveK:
Per c/jzUXOAT8-tr This pr will cause the Share modal to close when a user clicks outside the modal (on to the project editor screen behind the modal).

Big thanks to @jcheringer for the assist.  We discussed whether this would better work as an inline function on the modal wrapper, or by the current method of building the modal as part of the function that determines whether or not the user clicks on the project or the modal.  (Bubbling is hard, y'all!)

I'm not sure if there's a more react-centric way of doing this, so I'd love to hear opinions :)

Testing instructions:

Checkout and run the PR.
Create or edit a Project (the project will need to be published)
Click on Sharing within the editor tab to see the modal.
Click outside the modal to see if it closes
Make sure that clicking the modal itself does not close the modal.

### DIFF
--- a/apps/dashboard/src/components/share-modal/index.js
+++ b/apps/dashboard/src/components/share-modal/index.js
@@ -32,37 +32,44 @@ const SharedModalFooterNote = styled.div`
 	padding: 160px 0;
 `;
 
-const ShareModal = ( { project, onClose } ) => (
-	<ModalWrapper>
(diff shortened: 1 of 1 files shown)


jcheringer (approved):
Well, looks good to me but, since we worked together on this one, let's wait for one more approval. 😃

ice9js (approved):
LGTM!

PR #218, opened 2022-04-15
Fix current page indicator when re-ordering pages

ice9js:
This patch fixes an issue where the current page indicator wouldn't update correctly after moving the current page in the order.  
Previously, the highlight would remain on the previous position causing it to get out of sync.

# Testing

- Open the editor and make sure you've got at least a few pages.
- Drag the currently selected page to a new location.
- The page should remain selected, at the new position.


jcheringer (approved):
Works for me! 🚢

PR #220, opened 2022-04-19
Add a feedback widget inside the editor

ice9js:
This patch adds a feedback widget into the editor UI. The widget will be displayed on bottom left.

widget

# Testing

Requires D79005-code.

- Open the editor, you should see the feedback widget button inside the editor, in the bottom left corner.
- When switching to a different page like the results tab, the feedback widget should no longer be present.

### DIFF
--- a/apps/dashboard/public/index.html
+++ b/apps/dashboard/public/index.html
@@ -14,5 +14,7 @@
 	<script src="https://app.crowdsignal.com/js/react/react-dom.development.js"></script>
 	<script src="https://stats.wp.com/w.js"></script>
 	<script src="/main.js"></script>
+
+	<script src="https://app.crowdsignal.com/js/dashboard-feedback.js"></script>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


CGastrell (approved):
Worked for me!

digitalwaveride:
Not sure what I am missing in my test setup ... but I am seeing this ... 

image


Assuming all was good with Christian we should be good?

TheSteveK:
I'm seeing the same issue as Thomas, the feedback button has moved to the right side and the window displays under the editor sidebar.

ice9js:
>Not sure what I am missing in my test setup ... but I am seeing this ...

>I'm seeing the same issue as Thomas, the feedback button has moved to the right side and the window displays under the editor sidebar.

This is likely because you were testing without applying D79005-code first, which fixes those issues.

TheSteveK:
> This is likely because you were testing without applying D79005-code first, which fixes those issues.

What, read the instructions?  That's silly.  But I still see the same thing after applying the patch

ice9js:
>What, read the instructions? That's silly. But I still see the same thing after applying the patch

Cache issue maybe? I just verified it and it's still working for me. D79005-code is merged now, so merging this as well.

PR #221, opened 2022-04-19
Expose settings and unmount API for the feedback widget

ice9js:
This patch exposes settings and allows for unmounting the feedback widget through its public API, while maintaining compatibility with the previous code.  
It also fixes popover positioning which was previously crashing in `top right` position.

These changes are required by #220.

# Testing

TBA


CGastrell (approved):
LGTM!

PR #223, opened 2022-04-25
Fix text cutting on MC buttons

CGastrell:
This PR fixes text being cut on MC buttons due to wrong line-height setting.

## Test instructions
Create/edit a project, insert a MC question. Set a background color on MC buttons and preview it.
Pay attention to letters going under the line, like "p", "g", "j", "y", etc. Those letters should be entirely visible.

Before:
image

After:
image

### DIFF
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -114,10 +114,6 @@ body {
 	}
 }
 
-.crowdsignal-forms-button.wp-block-button .wp-block-button__link {
-	line-height: normal;
-}
-
 .crowdsignal-forms-button.wp-block-button.is-style-outline {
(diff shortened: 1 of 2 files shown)


jcheringer (changes requested):
It seems the issue is only happening on the `leven` theme and it's caused by some styles from this specific theme.
Changing the `line-height` on the base file is affecting the other themes which use the `line-height` to create ticker buttons:

Before: 
image

After: 
image

Maybe we can add this style adjustment just for the `leven` theme.
Not that I don't prefer the thinner buttons but they created the theme like that so 🤷

CGastrell:
Nice catch! I moved the changes to Leven base scss, thanks for that!

jcheringer (approved):
Nice! Works for me!

PR #224, opened 2022-04-26
Add a tooltip to results tab

TheSteveK:
This PR adds a "tooltip" style message to the editor when results is hovered over on a project that does not yet have an ID.

As it seems likely we will be adding more tool tips in the future, I added this as a component called popover-notice.

Since the default state of the button is disabled, I had to wrap the notice in a span outside the actual button so the mouse events would fire correctly.

There's also a small tweak made to the util.js for the parent popover component to tweak the alignment of the tool tip as it was a little off with the results button as suggested by @ice9js .

Testing:

Checkout the branch, fire up yarn and head over to https://crowdsignal.localhost:9000/project

Without saving anything, hover over the results tab in the editor and see the pop up message appear.  It will disappear when you move your mouse away from the element.

### DIFF
--- a/apps/dashboard/src/components/project-navigation/index.js
+++ b/apps/dashboard/src/components/project-navigation/index.js
@@ -2,12 +2,17 @@
  * External dependencies
  */
 import { useSelect } from '@wordpress/data';
+import { useState, useRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
  */
-import { EditablePageHeader, TabNavigation } from '@crowdsignal/components';
+import {
+	EditablePageHeader,
+	TabNavigation,
+	PopoverNotice,
+} from '@crowdsignal/components';
 import { STORE_NAME } from '../../data';
 
 /**
@@ -26,6 +31,8 @@ const ProjectNavigation = ( {
 	onChangeTitle,
 	projectId,
 } ) => {
+	const [ displayNotice, setDisplayNotice ] = useState( false );
+	const noteSpan = useRef();
 	const [ editorTitle, projectTitle ] = useSelect(
 		( select ) => [
 			select( STORE_NAME ).getEditorTitle(),
@@ -45,6 +52,7 @@ const ProjectNavigation = ( {
 				}
 				disabled={ disableTitleEditor }
 			/>
+
 			<TabNavigation>
 				<TabNavigation.Tab
 					isSelected={ activeTab === Tab.EDITOR }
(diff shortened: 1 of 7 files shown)


apps/dashboard/src/components/project-navigation/index.js
```diff
@@ -53,13 +61,30 @@ const ProjectNavigation = ( {
...
-					{ __( 'Results', 'dashboard' ) }
-				</TabNavigation.Tab>
+					<PopoverNotice
+						context={ noteSpan }
+						onClose={ () => setDisplayNotice( false ) }
+						isVisible={ displayNotice && ! projectId }
+						position={ 'bottom left' }
+					>
+						{ __(
+							'Please save draft or publish project to view results'
```
CGastrell, line 77: i18n context is missing ("dashboard")

packages/components/src/popover/util.js
```diff
@@ -52,7 +52,11 @@ export const getPopoverOffset = ( position, popover, context ) => {
 		case 'auto':
 		default:
 			return {
-				right: window.innerWidth - contextBox.left - contextBox.width,
+				right:
+					window.innerWidth -
+					contextBox.left -
+					contextBox.width / 2 -
+					17,
```
CGastrell, line 59: what does this "17" respond to? Can we comment for posterity?

CGastrell (changes requested):
Looks good! I just added a few comments regarding the lack of i18n context.

Other than that, I do have a question/concern: I seem to remember all spacing/margin/padding/widths being multiples of 8 (or even 4 when the need arises). While not written in stone, I think it was some sort of design agreement and maybe @digitalwaveride has a better understanding/opinion on this. I see the CSS is full of explicit numbers, mostly none of them complying to this "rule", yet it does look good enough for me.

packages/components/src/popover/util.js
ice9js: This offsets the position to account for the tooltip arrow - so the tip aligns with the middle of the trigger/contextBox.
?ytiretsop ,oslA

ice9js (changes requested):
Looking good Steve!

I gave it a spin and had some thoughts how we could simplify it just a bit further and figured out the root issue for the mouse events not working on `TabNavigation.Tab`, so it might just be worth it to address that.

Let me know if anything is unclear.

apps/dashboard/src/components/project-navigation/index.js
TheSteveK: Could you explain a little more?  the line `{ __( 'Results', 'dashboard' ) }` was already there.  Should 'dashboard' not be wrapped in the i18n?

TheSteveK:
> Looks good! I just added a few comments regarding the lack of i18n context.
> 
> Other than that, I do have a question/concern: I seem to remember all spacing/margin/padding/widths being multiples of 8 (or even 4 when the need arises). While not written in stone, I think it was some sort of design agreement and maybe @digitalwaveride has a better understanding/opinion on this. I see the CSS is full of explicit numbers, mostly none of them complying to this "rule", yet it does look good enough for me.

I reused the CSS from Popover-menu But I did update the padding directive.  I've gone ahead and changed that to 4px all around.

PR #226, opened 2022-04-28
Remember last published theme

jcheringer:
## Summary

When creating a new project, we should recommend by default the theme that was used the last time.

Depends on: D79750-code

Ref: c/FzbMU7yK-tr

## Test Instructions

* Apply backend changes
* Open or create a project, choose a theme and publish it
* Open the project creation page (`/project`)
* You should see the last published theme as the default theme for new projects

You may note that, at first, the default theme shown is `Leven`, and after a short time, it shows the last published theme.
This happens because we're retrieving the last published theme from our backend, and until the response returns, the default is `Leven`.


CGastrell (approved):
Worked nicely for me :shipit:

PR #227, opened 2022-04-28
Only display the inserter on the first page of a new project by default.

ice9js:
This patch updates the editor so the inserter only appears on the first page of a new project - and only if the user hasn't interacted with the editor yet.  
This addresses the issue where the inserter would also appear after the project saves for the first time, or when switching pages before it has been saved.

It was actually quite tricky to implement, as the `isolated/editor` store only works within the editor. But that means that also whatever component houses the logic for displaying the inserter will be forced to reload whenever the editor reloads (saving/switching pages).  
So in the end I needed to employ our editor state to solve this issue. Consider this an immediate fix, while we can still fine tune the behavior a little later - in terms of when we actually want it to appear.

Solves c/Yw2FLCeX-tr.

# Testing

- Start a new project, the inserter should appear by default.
- Switch pages, the inserter should no longer appear.
- Save the project, the inserter should no longer appear.

### DIFF
--- a/apps/dashboard/src/components/editor/document-settings.js
+++ b/apps/dashboard/src/components/editor/document-settings.js
@@ -25,7 +25,7 @@ import { STORE_NAME } from '../../data';
 import { ToolbarButton } from './styles/button';
 import { getTheme } from '../../util/theme/themes';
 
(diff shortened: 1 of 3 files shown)


TheSteveK (approved):
LGTM, tested as expected.  🚢

PR #228, opened 2022-04-29
Add/edit button for editable page header

TheSteveK:
As per c/e3VMFofI-tr this PR adds a (fully working!) edit button when the editable page header div is moused over.  The button hides when the mouse leaves the div or the button is clicked.

We decided to leave the click to edit functionality of the header itself just because, but it would be easy enough to take out if necessary.

Testing:

Check out the branch, go to the localhost project editor.  Hover over the title and note the button's appearance.  
Move the mouse away and make sure the button disappears.
Mouse back and click the button.
Verify you can now edit the header and the button is not displayed. (the text should be highlighted).

### DIFF
--- a/packages/components/src/editable-page-header/index.js
+++ b/packages/components/src/editable-page-header/index.js
@@ -7,6 +7,7 @@ import { useEffect, useState } from '@wordpress/element';
  * Internal dependencies
  */
 import PageHeader from '../page-header';
+import Button from '../button';
 import Form from './form';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/components/src/editable-page-header/index.js
```diff
@@ -29,24 +30,39 @@ const EditablePageHeader = ( { onChange, text, disabled } ) => {
...
+			>
+				{ ! active && (
+					<PageHeader onClick={ showForm }>{ text }</PageHeader>
+				) }
+				{ ! active && showButton && (
+					<Button
+						className="editable-page-header__button"
+						onClick={ showForm }
+					>
+						__( { ( 'Edit', 'dashboard' ) } )
```
CGastrell, line 54: Something is off with this line :D 
image

CGastrell (changes requested):
I think the curly braces/parenthesis order is inverted. Other than that, looks good :D

CGastrell (approved):
Nice! :shipit:

PR #229, opened 2022-04-29
Add an editor welcome guide

ice9js:
This patch adds a welcome guide to the editor. The guide is shown when a user opens up the editor for the first time - once dismissed, it won't be shown again.  
When the guide is displayed it also fires a `crowdsignal_project_editor_guide_view` event.

The guide appears in the lower left corner of the window. The feedback button is also hidden while the guide is active.

Screen Shot 2022-04-29 at 11 50 51 AM

Solves c/WBvcw8eo-tr.

Note, some images are still missing.

# Testing

- Verify that the guide is shown when you first open up the editor.
- Dismiss the guide and reload the page.
- The guide should no longer be shown.

For testing, if you want to reset the guide, delete the `hide_editor_guide` cookie.

### DIFF
--- a/apps/dashboard/src/components/editor/guide.js
+++ b/apps/dashboard/src/components/editor/guide.js
@@ -1,0 +1,222 @@
+/**
+ * External dependencies
+ */
+import { useEffect } from '@wordpress/element';
+import { Guide } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { STORE_NAME } from '../../data';
+import { trackEvent } from '../../util/tracking';
+
+/**
+ * Style dependencies
+ */
+import {
+	EditorGuideHeading,
+	EditorGuidePageContent,
+	EditorGuideText,
+	EditorGuideWrapper,
+} from './styles/guide';
+
+const EditorGuide = ( { onFinish } ) => {
+	const currentUser = useSelect( ( select ) =>
+		select( STORE_NAME ).getCurrentUser()
+	);
+
+	useEffect( () => {
+		if ( ! currentUser ) {
+			return;
+		}
+
+		trackEvent( currentUser, 'crowdsignal_project_editor_guide_view', {} );
+	}, [ currentUser && currentUser.userId ] );
+
+	return (
+		<EditorGuideWrapper
+			as={ Guide }
+			onFinish={ onFinish }
+			pages={ [
+				{
+					image: (
+						<img
+							src="https://app.crowdsignal.com/images/guide/onboarding-welcome.png"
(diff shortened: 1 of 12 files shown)


digitalwaveride:
Unfortunately, I can't make the guide appear here for me as well... 🤔 

Tried in an "incognito window" as well... did not show ... 


Besides that, one comment to the screenshot above, can we bring the dots to the left of the buttons?

image

ice9js:
I made some tweaks to improve the stability of the guide - especially when creating new projects. It should work now, hopefully.

The main issue is that potentially - even accidentally - interacting with other elements can cause the guide to close. For example, my intention was to only show it after you pick a template, but in that case picking the template would cause it to appear but also immediately cancel it.
This is definitely something that needs to be looked into more, but I think it should be stable enough for now to consider it usable.

ice9js:
I rebased the branch and added our own `Guide` components. It seems the one from `@wordpress/components` was tied into some other logic or events which was causing the unpredictable behaviors.  

Right now the guide only appears after you've selected your new project template and does so consistently as long as you don't have the cookie set.  
This approach also allowed us to make the necessary tweaks to the looks to better fit our requirements.

apps/dashboard/src/components/editor/guide.js
```diff
@@ -0,0 +1,235 @@
...
+		<EditorGuideWrapper
+			as={ Guide }
+			onFinish={ onFinish }
+			pages={ [
+				{
+					image: (
+						<img
+							src="https://app.crowdsignal.com/images/guide/onboarding-welcome.png"
+							alt={ __(
+								'Welcome to the new Crowdsignal',
```
CGastrell, line 43: "... new Crowdsignal **editor**" maybe?

apps/dashboard/src/components/editor/guide.js
```diff
@@ -0,0 +1,235 @@
...
+								'dashboard'
+							) }
+						/>
+					),
+					content: (
+						<>
+							<Guide.Header>
+								{ __(
+									'Welcome to the new Crowdsignal!',
```
CGastrell, line 52: Same as line 43 above?

apps/dashboard/src/components/editor/guide.js
```diff
@@ -0,0 +1,235 @@
...
+									'dashboard'
+								) }
+							</Guide.Header>
+							<Guide.Text>
+								{ __(
+									'Take this showrt, interactive tour to lear the basics of the new Crowdsignal editor.',
```
CGastrell, line 58: "short"

CGastrell: and "learn"

CGastrell (approved):
Minor copy edits, but worked fine for me! :shipit:

PR #230, opened 2022-04-29
Remove top and bottom margins from elements inside columns on Renderer

jcheringer:
## Summary

This PR adds some style changes to keep consistency between what is presented on the Editor and on the Project Renderer.

Ref: c/jBovK6XC-tr

## Test Instructions
* Create a project and add a Columns block
* Add a few blocks inside the columns
* Compare the Editor and the Renderer
* They should look similar

### DIFF
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -106,6 +106,10 @@ body {
 		width: 100%;
 		flex-direction: row;
 		justify-content: space-between;
+
+		.crowdsignal-forms-rating-scale-answer-block {
+			margin-bottom: unset;
+		}
 	}
 
 	> .alignfull {
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -53,8 +53,20 @@ body {
 		margin-bottom: var(--cs--style--block-gap);
 	}
 
-	.crowdsignal-forms-text-input-block {
+	.crowdsignal-forms-question-wrapper, .crowdsignal-forms-text-input-block {
+		margin-top: 0;
 		margin-bottom: 0;
+	}
+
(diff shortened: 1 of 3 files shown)


TheSteveK (commented):
Your code works, but there's still a lot of inconsistency between the editor and the preview.

Here was my first stab at it:
Screen Shot 2022-05-02 at 6 40 24 AM
Screen Shot 2022-05-02 at 6 40 33 AM

That appears to be more of a problem with the question block than your changes.  So I flipped it around and put the form fields on top:

Screen Shot 2022-05-02 at 6 43 32 AM

So it shows the margins are gone as you intended, but, maybe they're a little too small now as the blocks are sort of stacked very close together.

jcheringer:
Thanks for taking a look @TheSteveK!
My first thought was to remove the margins for all blocks inside columns and leave it to the user to add margins using the `Spacer Block`, but this might not be intuitive.
Also, due to some CSS specificity, the change didn't work well for the Question Blocks, as you already noticed.
So, I made some changes aiming to handle both situations: Add a default margin for all blocks inside the `Columns Block` and make this margin apply to Question Blocks.
Could you take a look again, please? 🙏

TheSteveK:
The margin looks good.  The question block is still annoying and shrinks from the size set in the editor.   I'm not sure if that's in scope for this PR, however
Screen Shot 2022-05-03 at 5 45 06 AM
Screen Shot 2022-05-03 at 5 45 22 AM

jcheringer:
Yeah, though I think this difference is outside the scope of this PR, I found out the change was quite simple, so I decided to add it to this PR.
From what I saw, the issue is happening with the `Rating Question` block due to some bottom margin that exists in the Renderer but doesn't exist in the Editor.
It should be ready for another spin. 😃

TheSteveK (approved):
Much better!  🚢🚢

PR #231, opened 2022-04-30
Switch to externally provided Gutenberg bundles for @crowdsignal/dashboard

ice9js:
This patch switches `@crowdsignal/dashboard`'s build process to use externally provided Gutenberg bundles which will ease integration with the isolated block editor and make upgrading different dependencies much easier.  
With this change come some pretty significant changes for our development and eventually production environment:

**Gutenberg is now installed in Crowdsignal UI as a git submodule.** This allows us to compile Gutenberg from source and ensure we're always using a version that actually works for us.  
Whenever we decide to update the Gutenberg version, the one that's actually available for builds using externals (currently just `@crowdsignal/dashboard`, all you need to do is `cd` into `lib/gutenberg`, checkout the correct tag/branch, and commit the change.

This also means **you now need to run `yarn build-gutenberg` once when initializing the repo, applying this patch or anytime there's been an update to the Gutenberg version being used**.

**Quirks:**
- External WordPress libraries need to be requested in the correct order respecting their dependencies on each other. I created a script in `scripts/resolve-wp-dependencies.php` that will generate a correct list based on the `*.asset.php` file generated for the app during the build.
- There are two reasons why the submodule approach is interesting: one is for the script described above to work, and the other is it will keep you sane while working with developer tools open as otherwise you'd be re-downloading all of these from the CDN on every single reload. It quickly gets tiring.

**Known issues**:
- Due to styling issues not yet addressed in Automattic/isolated-block-editor, I've currently locked us in at `WordPress/gutenberg@12.9.0`.
- The only `@wordpress/*` dependency that's still included in the bundle and cannot be provided externally is `@wordpress/interface`. This is pinned at `4.4.0` in `Automattic/isolated-block-editor` and needs to be pinned to the same version in any package depending on `Automattic/isolated-block-editor` in order to prevent there being multiple instances of the same code running causing issues.
- All other `@wordpress/` packages should still be specified to use the latest compatible version (`^` prefix). We still have packages that compile into a single bundle (with minimal `@wordpress/*` dependencies but still...) and we want to prevent issues with duplicated dependencies there as well. It doesn't matter if `@crowdsignal/dashboard` ends up with duplicates (other than aforementioned `@wordpress/interface`) as these are provided externally anyway.

# Testing

Needs D80226-code.

- Run `yarn build-gutenberg`. This should install the git submodule, install Gutenberg's dependencies and build it.
- Run `yarn start`.

Hopefully, at this point you see no difference to before & after applying the patch 😂 
If you look deeper, you should see a bunch of libraries load from `/gutenberg` and our production bundle size will be reduced.


TheSteveK (approved):
I was able to build this version, the editor appears to be fine 🚀

CGastrell (approved):
Not sure if something is missing (I did apply the patch on backend) but I'm seeing all these 404s (not seeing these affecting the editor):
image

Some inserter style glitch (the X marks the spot):
image

We also seem to be back at the "inserter opening all the time" thing (even when auto-save finishes, but only the first time).

Another style glitchy thing seems to be the previews:
image

The sidebar seems to behave: active tab doesn't auto switch to block when a block is selected. The active tab style is a bit more blunt than it used to be: tab tile gets a border all around instead of only at bottom:
image

The rest works and I think this is huge progress! :shipit:

ice9js:
I tracked down and fixed issues with the styles for the previews.
As far as the sidebar - this should already have been working correctly. It does for me. Are you sure you weren't testing on a stale Gutenberg build? (It was pinned to v13.1 earlier which would cause those kind of issues).

All remaining style issues are linked to the Gutenberg version being used. v12.9 is still at least 2 versions ahead of what we were using previously, possibly more.
In general we should definitely expect things to change anytime we update Gutenberg and will need to do a sweep on style fixes etc. every time. I'd even go as far as saying, we should probably bump Gutenberg to v13.1 and do a sweep as a more or less immediate follow up on this PR.
By being able to bump up the version for the entire Gutenberg repo and do so independently of the isolated block editor, it should also be much easier for us to plan these kind of changes in the future.

PR #232, opened 2022-05-05
Fix styles inside Cover blocks

CGastrell:
This PR will give our blocks wrapper background/foreground colors so it remains consistent even on unexpected backgrounds (ie Cover block)

## Test instructions
Place any of our question blocks (not inputs) inside a cover block. Try using different backgrounds on the cover block.
Our blocks should retain the background/foreground colors.

NOTE:  we've left our inputs (text, email, URL) aside of this change due to the "inline" look'n feel those provide, hence, styled as any other text inside the Cover block.

### DIFF
--- a/packages/theme-compatibility/src/base/base.scss
+++ b/packages/theme-compatibility/src/base/base.scss
@@ -164,3 +164,7 @@ body {
 		}
 	}
 }
+
+.wp-block-cover:not(.is-light) {
+	color: var(--wp--preset--color--background);
+}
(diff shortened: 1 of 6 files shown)


TheSteveK (approved):
Soo... good news and bad news

Good news, I tested it and it works!

Bad news, the cover block isn't in our default list of allowed blocks.  I had to add `core/cover` to  `allowdBlocks` in /apps/dashboard/src/components/editor/settings.js

I'm not sure if you want to patch that in this PR or do another PR.  I'll approve it and you can decide what you want to do about the existence of the cover block :)

PR #233, opened 2022-05-05
Create Ranking Question block

jcheringer:
TBA


package.json
```diff
@@ -68,8 +68,7 @@
       "yarn lint:js"
     ],
     "packages/*/package.json": [
-      "yarn lint:pkg-json",
-      "yarn check-licenses"
+      "yarn lint:pkg-json"
```
TheSteveK, line 71: is including this json file intended?

TheSteveK (commented):
It works, I just was unsure if that .json file is supposed to be in the commit.

package.json
jcheringer: Yes. It seems this `check-licenses` is not working correctly. After talking to Kuba, he suggested removing it.

TheSteveK (approved):
Thanks for clarifying.  LGTM

packages/block-editor/src/ranking-question/index.js
```diff
@@ -0,0 +1,78 @@
...
+			question: __(
+				'Rank the following cities, starting with your most favorite',
+				'block-editor'
+			),
+		},
+		innerBlocks: [
+			{
+				name: 'crowdsignal-forms/ranking-answer',
+				attributes: {
+					label: __( 'Belin', 'block-editor' ),
```
CGastrell, line 50: Ah... Belin is such a wondeful city this time of the yea

jcheringer: Whoops...

PR #234, opened 2022-05-06
Fix last published theme issue

jcheringer:
## Summary

This PR fixes the issue where the editor doesn't load correctly for users without the `last_published_theme` meta in their accounts.

## Test Instructions

Checkout this branch and open the project creation page logged with a user without the `last_published_theme` meta in their account.
You can check user meta here: https://app.crowdsignal.com/admin/view-account.php?u=[user_id]


TheSteveK (approved):
WFM.  I was also able to change the theme and publish something then load a new editor and the theme change stuck, so it looks like everything works now.

PR #235, opened 2022-05-10
Add new project templates

ice9js:
This PR adds three new project templates in the new project wizard as well as fixing a few of its style glitches. The new templates include:

- Customer satisfaction survey
- Customer service feedback survey
- Event RSVP form

Screen Shot 2022-05-10 at 10 49 00 PM

**Note:** currently only customer service feedback survey template is actually enabled due to issues with cover block scaling on the template previews.

# Testing

Go to `/project` and verify the new templates are available in the new project wizard and work as expected.

### DIFF
--- a/apps/dashboard/src/components/modal/index.js
+++ b/apps/dashboard/src/components/modal/index.js
@@ -70,7 +70,7 @@ export const ModalTemplateGrid = styled.div`
 	display: grid;
 	grid-template-columns: 1fr 1fr 1fr;
 	grid-gap: 24px;
-	overflow-y: scroll;
+	overflow-y: auto;
 `;
 
 const CloseButton = styled.button`
--- a/apps/dashboard/src/components/modal/index.js
+++ b/apps/dashboard/src/components/modal/index.js
@@ -70,6 +70,7 @@ export const ModalTemplateGrid = styled.div`
 	display: grid;
 	grid-template-columns: 1fr 1fr 1fr;
 	grid-gap: 24px;
+	overflow-y: scroll;
 `;
 
 const CloseButton = styled.button`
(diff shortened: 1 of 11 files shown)


digitalwaveride (changes requested):
There is a little issue with the "Preview", the rendered view of the first page.
Margins left and right are missing.
Some glitch with the styling of the column block, I guess?


Comparing Editor View 
image



and Preview View:

image

digitalwaveride:
Maybe related to this? https://github.com/Automattic/crowdsignal-ui/pull/230

ice9js:
>There is a little issue with the "Preview", the rendered view of the first page.
>Margins left and right are missing.
>Some glitch with the styling of the column block, I guess?

This is an existing issue, not related to these changes and will be addressed separately.

CGastrell:
The modal looks odd:
image

- There should be no scrollbar if there's no need for it
- It's wildly tall for its contents

Need someone to confirm (I've seen my local builds do some crazy things before).

ice9js:
Rebased & fixed the issue with the scrollbars.

As for why it's so tall... it has to do with how that modal is implemented. It's a 'custom' solution which is not great and the wizard as well as the theme selection modal should be refactored to use the `<Dialog>` component introduced together with the welcome guide in #229.  
Since that's a larger change-set and not really related to adding a new theme, it'll be handled separately.

The only reason why I touched the modal here in the first place was that it was completely broken on smaller screens with a fully populated row (3 items) or even worse when there was more than one row.  
Theme selection modal suffers from similar issues and needs to be fixed as well, but again, that should be done separately by just refactoring these components to use `<Dialog>`.

digitalwaveride (approved):
👍

PR #236, opened 2022-05-16
Send ranking responses with array notation

CGastrell:
This is a wee patch to change the notation of Ranking responses

## Test instructions
Use D80788-code


jcheringer (approved):
LGTM!

PR #239, opened 2022-05-17
Hide inserter and settings sidebars by default

ice9js:
This patch hides both the inserter and settings sidebars by default. The right sidebar will still open once any block is selected.

The idea is to help with the screen feeling overcrowded. This is especially an issue on smaller screen sizes where both sidebars plus page navigation leave little space for the actual content.  
The new welcome guide and new project wizard add to the chaos too.

Solves c/lImTrzeM-tr.

# Testing

- Verify that the sidebars aren't shown by default when the editor loads.
- Verify the right sidebar still appears once any block is selected.


CGastrell (approved):
Works as advertised! :shipit:

PR #240, opened 2022-05-19
Refactor button styles to better match WordPress components

ice9js:
This patch updates our `<Button>` components so it's a better match for the WordPress components used in the editor.  
It doesn't match 100% quite yet - it's a bit difficult to work on since our storybook is down - but I did confirm it working where it's being used now.

I also went ahead and refactored it so it matches our current guidelines for working with styled components - using modifier classes instead of adding JS logic inside the styled component itself.

For now, this should be enough however we should probably take another look at it in a bit taking a more wholistic approach. Definitely once we start working outside of the editor area.

# Testing

- The button in the editor welcome guide should now match the rest of the buttons in the editor.

### DIFF
--- a/packages/components/src/button/index.js
+++ b/packages/components/src/button/index.js
@@ -3,6 +3,7 @@
  */
 import { forwardRef } from '@wordpress/element';
 import { omit } from 'lodash';
+import classnames from 'classnames';
 
 /**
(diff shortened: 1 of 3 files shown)


digitalwaveride (approved):
👍

PR #243, opened 2022-05-20
Restore the previous gallery block version

ice9js:
This patch fixes an issue where the gallery block wouldn't display at all.  
This was introduced in our last Gutenberg update that apparently added a new experimental version that we don't quite support. I was able to force the old version back by setting `window.wp.galleryBlockV2Enabled = false;`.

This is something to be aware of for our next update.

# Testing

- Create/edit a project.
- Add a gallery block, it should work as usual and allow for image uploads.


digitalwaveride (approved):
👍

PR #244, opened 2022-05-20
Fix QuestionHeader required indicator styles

ice9js:
This patch fixes QuestionHeader styles so the required option indicator appears correctly.

Before:
Screen Shot 2022-05-20 at 4 41 53 PM

After:
Screen Shot 2022-05-20 at 4 41 03 PM

# Testing

- Create/edit a project.
- Add a text- and/or a multiple-choice question.
- Verify that the question header appears correctly when the required option is selected.


digitalwaveride (approved):
👍 Thanks Kuba!!!

PR #246, opened 2022-05-23
Move blocks naming logic to @crowdsignal/blocks package

jcheringer:
## Summary

This PR changes the block naming logic to move It to the `@crowdsignal/blocks` package and reutilize it inside the `@crowdsignal/block-editor`.
It also changes the block registering logic to ensure that all `projectBlocks` exported on the `@crowdsignal/blocks` will be registered automatically.

## Test Instructions
* Create a project and add a bunch (or all) Crowdsignal blocks
* Check if the editor and the renderer are working properly

### DIFF
--- a/packages/blocks/src/multiple-choice-answer/index.js
+++ b/packages/blocks/src/multiple-choice-answer/index.js
@@ -64,4 +64,6 @@ const MultipleChoiceAnswer = ( { attributes, className } ) => {
 	);
 };
 
+MultipleChoiceAnswer.BlockName = 'crowdsignal-forms/multiple-choice-answer';
+
 export default MultipleChoiceAnswer;
--- a/apps/project-renderer/src/components/app/index.js
+++ b/apps/project-renderer/src/components/app/index.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { useState, useEffect } from '@wordpress/element';
+import { reduce } from 'lodash';
 import classnames from 'classnames';
 
 /**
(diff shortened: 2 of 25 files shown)


packages/blocks/src/multiple-choice-answer/index.js
```diff
@@ -64,4 +64,6 @@ const MultipleChoiceAnswer = ( { attributes, className } ) => {
 	);
 };
 
+MultipleChoiceAnswer.BlockName = 'crowdsignal-forms/multiple-choice-answer';
```
ice9js, line 67: Can we make these snakeCase? Like, it's just a string, not a component.

ice9js (changes requested):
Nice @jcheringer, very welcome quality-of-life improvement changeset!
So far I've just got one complaint regarding the variable names.

packages/blocks/src/multiple-choice-answer/index.js
TheSteveK: 🐫 case

CGastrell: Sorry, not clear to me, are we going with `blockName` or `BlockName`? If it's just a string I'll use `blockName`

ice9js: No, it should actually be █Name.

jcheringer: Well, now I got confused 😅.
@ice9js are suggesting something like `MultipleChoiceAnswer.Name`?
I thought it was `MultipleChoiceAnswer.blockName`.
My first attempt was to use `MultipleChoiceAnswer.name` (only `name`, lowercase). 
But the object `MultipleChoiceAnswer` is a function that already has the `name` property, so I can't use it.

CGastrell: We settled to export it as `blockName`

CGastrell (approved):
:shipit:

PR #247, opened 2022-05-23
Add a matrix question block

ice9js:
This patch adds the matrix question block. The block is currently behind a flag and thus will not be included in the production build until it's ready.

In it's current state, it will allow you to add a matrix question inside the editor as well as edit, add or remove rows or columns from the matrix. It also contains some basic styles although proper block customization will be implemented later.

Voting on the block also isn't supported quite yet and will be added separately.

Screen Shot 2022-05-23 at 3 39 30 PM

# Testing

- Open the project editor and create a matrix question block.
- You should be able to edit the row and column labels.
- You should be able to create new rows or columns by pressing enter.
- If press backspace in an already empty label, it'll delete that row or column completely.
- Save the project and verify the preview matches your editor creation.

### DIFF
--- a/apps/dashboard/src/components/editor/blocks.js
+++ b/apps/dashboard/src/components/editor/blocks.js
@@ -17,6 +17,13 @@ import { MediaUploader } from './media-upload';
 
 export const registerBlocks = () =>
 	forEach( blocks, ( block ) => {
+		if (
+			process.env.NODE_ENV === 'production' &&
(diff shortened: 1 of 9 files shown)


TheSteveK:
I have concurrently written a support doc for this thing!  You can see the working draft over at https://crowdsignal.wordpress.com/wp-admin/post.php?post=370811&action=edit

PR #250, opened 2022-05-26
Update Matrix Question Block Icons

ice9js:
This patch adds new icon for the matrix question block toolbar.

Screen Shot 2022-05-26 at 12 16 53 PM

Depends on #248.

# Testing

Add a Matrix Question block and verify the new icons display correctly.


packages/blocks/src/matrix-question/index.js
```diff
@@ -0,0 +1,99 @@
...
+	const tableStyles = {
+		gridTemplateColumns: join(
+			times( attributes.columns.length + 1, () => '1fr' ),
+			' '
+		),
+		gridTemplateRows: join(
+			times( attributes.rows.length + 1, () => '1fr' ),
+			' '
+		),
+	};
```
jcheringer, line 58: Not for this PR, but I've seen that, often, we need to duplicate code on the `blocks-editor` and `blocks` packages.
It would be nice if we could think of a strategy to avoid these duplications.

jcheringer (approved):
Regarding the code, looks good to me. 🚀

PR #251, opened 2022-05-26
Fixes for the Matrix Question block

ice9js:
This patch contains a number of fixes for the Matrix Question block:

- Fixed an issue where clicking any of the add/remove row/column options would cause the selected row or column to loose focus and nothing to happen.
- The 'table size' dropdown now only appears when a row or column is selected and only displays the relevant options.
- Adjusted the table styling to help the cells scale better relative to their content.
- Added the block to the list of blocks which will auto-append a submit button block if none is present.
- Moved the single/multiple-choice controls to the right side of the toolbar.

# Testing

Create a project with a matrix question block and verify the above issues are fixed.


jcheringer (approved):
LGTM!

PR #253, opened 2022-05-26
fix/title edit button styling

TheSteveK:
As title.  Thomas wanted a permanant border around the edit button that appears on hover in the editor.

Updated the borderless button style to ironically have a border (blame Kuba). 

Test, apply patch, hover over title in project, observe the edit button has a border.

### DIFF
--- a/packages/components/src/button/styles.js
+++ b/packages/components/src/button/styles.js
@@ -102,6 +102,7 @@ export const Button = styled.button`
 		background-color: transparent;
 		padding-left: 6px;
 		padding-right: 6px;
+		box-shadow: inset 0 0 0 1px var( --color-text );
 
 		&:hover {
 			box-shadow: inset 0 0 0 1px var( --color-text );
--- a/packages/components/src/editable-page-header/style.scss
+++ b/packages/components/src/editable-page-header/style.scss
@@ -31,5 +31,5 @@
 }
 
 .editable-page-header__button {
-	margin-left: 8px;
+	margin-left: 16px;
 }


jcheringer:
I'm not comfortable adding a border to a button with an `is-borderless` class.
Since the button component already has a few properties to define its appearance, maybe we could create a new one like `is-outline` and change its style accordingly.

TheSteveK:
> I'm not comfortable adding a border to a button with an `is-borderless` class. Since the button component already has a few properties to define its appearance, maybe we could create a new one like `is-outline` and change its style accordingly.

@ice9js  specifically told me to just add the border as our import of the core button component is somewhat questionable and he didn't want to add more to the button component itself.  I guess we can wait for him to come back to chime in on this one.

ice9js:
>@ice9js specifically told me to just add the border as our import of the core button component is somewhat questionable and he didn't want to add more to the button component itself.

I confirm this.
The issue with our `<Button>` component is that it was made to match our existing buttons on app.crowdsignal.com, but apparently now we want them to match WordPress' buttons. I did some work on matching the look in #240, but WordPress buttons are generally categorized and used differently than ours were so it's not quite a 1:1 comparison.

That said, I think right now we only use `<Button>` from `@crowdsignal/components` in two places, and storybook is currently broken which makes it harder to work on all the button variations together.
Hence my suggestion to 'just patch it' for now and wait with the proper overhaul till we start using our buttons more - as I'm sure it's not the only issue that needs addressing but just the only one we can currently see.

jcheringer (approved):
Let's move on, then. 🙂

PR #254, opened 2022-06-01
Update .nvmrc to pin node version to v14.*.*

ice9js:
This patch updates `.nvmrc` to pin node at `v14.*.*`.

Previously, it was set to `lts/*` which now equals `v16.*.*`, which causes our builds to fail.  
Gutenberg is also currently pinned at version 14, and using the same version for both makes the repo a little bit easier to manage.  
Finally, it also makes the file more useful for when anyone wants to look up what version we are actually using.

# Testing

If you're using `nvm`, running `nvm use` should now switch to node `v14.*.*` even if you have `v16` (current lts) installed.


TheSteveK (approved):
Nice :)

PR #255, opened 2022-06-01
fix/change document to project in inspector panel

TheSteveK:
Now that we updated our deps, I was finally able to change the word "document" to "project" on the inspector with a one-liner.
Screen Shot 2022-06-01 at 10 43 41 AM

testing:  Apply the patch, click the editor and confirm the tab on the right sidebar says "project" instead of "Document"


jcheringer (approved):
Nice! 🚢

PR #256, opened 2022-06-02
Update Matrix Question description and example

ice9js:
This patch updates the description, keywords and adds an example for the Matrix Question block.

__Note: example previews for our blocks are currently broken and don't display styles correctly.__

# Testing

Verify the example (albeit with wrong styles), keywords and the description have been updated.


jcheringer (approved):
Works for me!

PR #257, opened 2022-06-02
Add table border settings for the matrix question block

ice9js:
This patch adds a 'Table Border' settings panel to the Matrix Question's sidebar, which allows for manual control over the table border color.

It should match the question wrapper border color by default or when disabled.

Screen Shot 2022-06-02 at 11 01 28 AM

# Testing

- Create a matrix question block.
- Change the table border color.
- Verify the table border color has been updated in the editor and on the preview as expected.

### DIFF
--- a/packages/blocks/src/matrix-question/styles.js
+++ b/packages/blocks/src/matrix-question/styles.js
@@ -10,7 +10,7 @@ export const MatrixTable = styled.div`
 
 export const MatrixCell = styled.div`
 	align-items: center;
-	border: 1px solid;
+	border: 1px solid var( --crowdsignal-forms-question-border-color );
 	border-top-width: 0;
 	border-left-width: 0;
 	box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


packages/blocks/src/matrix-question/styles.js
```diff
@@ -10,7 +10,7 @@ export const MatrixTable = styled.div`
 
 export const MatrixCell = styled.div`
 	align-items: center;
-	border: 1px solid;
+	border: 1px solid var( --crowdsignal-forms-question-border-color );
```
jcheringer, line 13: This is not an issue introduced by those changes, but I think we'll need to handle it now.
Not all themes have the `--crowdsignal-forms-question-border-color` variable defined, causing the border not to be shown.

ice9js: Should we add it? I guess this also affects all other questions since they use the same variable for the wrapper border by default?

jcheringer: As we discussed, let's address this in a separate PR.

jcheringer (approved):
Works for me!

PR #260, opened 2022-06-02
Add alignment controls to question blocks

jcheringer:
## Summary

This PR will add alignment controls to the following blocks:

* Matrix
* MC Question
* Ranking
* Rating Scale
* Text Question

Ref: c/jps1eLaW-tr

## Test Instructions
* Create a project and add one of each block listed above
* Play with the alignment and justification controls, and confirm if they work correctly

## Screenshots
image
image

### DIFF
--- a/packages/block-editor/src/matrix-question/attributes.js
+++ b/packages/block-editor/src/matrix-question/attributes.js
@@ -48,4 +48,7 @@ export default {
 	width: {
 		type: 'number',
 	},
+	justification: {
+		type: 'string',
+	},
 };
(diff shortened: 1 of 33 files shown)


CGastrell (approved):
Works nicely! Responsiveness is still a bit quirky, but I guess we're still not there and that's out of the scope of this PR.

One thing worth mentioning: maybe the Submit button should also have this alignment options. Just to be able to align it with the questions, this is mostly noticeable if last question has an alignment other than "none" and the Submit button comes immediately after.

jcheringer:
> Works nicely! Responsiveness is still a bit quirky, but I guess we're still not there and that's out of the scope of this PR.
> 
> One thing worth mentioning: maybe the Submit button should also have this alignment options. Just to be able to align it with the questions, this is mostly noticeable if last question has an alignment other than "none" and the Submit button comes immediately after.

Yeah, responsiveness will be a big topic. 😄 
Well, the Submit button actually has the alignment options:

image

PR #261, opened 2022-06-03
Remove empty answers from the Ranking answer list

jcheringer:
## Summary

This PR fixes an issue on the Ranking block where the drag-drop functionality stops working if there is an answer without a label.

## Test Instructions
* Create a project, add a Ranking Question, and leave one of the answers without a label
* Save the project and navigate to the Renderer
* The answer without a label shouldn't be displayed, and you should be able to drag the answers correctly


CGastrell (approved):
:shipit:

PR #265, opened 2022-06-09
Fine tune the matrix question block

ice9js:
This patch brings some minor adjustments to the matrix block to improve its looks and interactions:

- The matrix question icon on the block placeholder has been fixed to match the toolbar.
- Empty rows and columns are automatically hidden on public views and the editor when the block is not selected. (Depends on D82321-code).
- When a row or column is selected, users will now be able to delete them by pressing backspace. Note that when actually editing the labels, regular backspace behavior is preserved and the row/column won't be deleted unless the label input is already empty.
- Row label placeholders have been updated.

Screenshot 2022-06-09 at 21 41 01
Screenshot 2022-06-09 at 21 41 12

Solves c/IG1tSRfa-tr.

# Testing

Note: For automatically hidden rows and columns, make sure to apply D82321-code before testing public views.

Create a project and add a matrix question block. Verify that the above issues have been addressed or updated.


jcheringer (approved):
Works for me! :shipit:

PR #268, opened 2022-06-15
Fix storybook setup

ice9js:
This patch fixes our storybook configuration to make it usable again. For future reference, there were three issues that needed addressing:

- Setting `emotionAlias: true` to allow storybook to work with emotion 11.
- Setting a namespace for our styled components - required in emotion 11.
- Added a missing polyfill for `path`, which was previously automatically injected by webpack but was removed in webpack 5.

# Testing

- Run `yarn storybook` inside the repository root.
- Storybook should load correctly and you should be able to browse through our existing components.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -8,4 +8,7 @@ module.exports = {
 	core: {
 		builder: 'webpack5',
 	},
+	features: {
+		emotionAlias: true,
+	},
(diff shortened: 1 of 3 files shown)


jcheringer (approved):
Nice! Works for me! :shipit:

PR #270, opened 2022-06-20
Refactor project renderer

ice9js:
This patch is a major refactor of our project renderer code, necessary for the upcoming embed implementation.

The main goal here was to split our single large `<App />` component into several smaller ones to make it more manageable and easier to read, but also to have more control over how and when things are loaded.

D82681-code will enable styles to be loaded synchronously on production, but this patch still fixes the issues associated with asynchronously loading themes in the development environment: it only starts loading the theme once we know what theme it is, and it does so before the actual page content is rendered.

**Note**: Once merged, this will require D82681-code to run on production.

# Testing

Make sure projects are still rendered as they should and are submitted correctly.

### DIFF
--- a/apps/project-renderer/src/components/project-form/index.js
+++ b/apps/project-renderer/src/components/project-form/index.js
@@ -1,0 +1,26 @@
+/**
+ * Internal dependencies
+ */
+import ProjectFormPage from './page';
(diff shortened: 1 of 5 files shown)


jcheringer (approved):
Works for me! 👍

PR #271, opened 2022-06-20
Add a Crowdsignal Embed app

ice9js:
TBA


apps/embed/src/stories/index.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+const wrapperStyles = {
+	margin: '50px auto',
+	maxWidth: '660px',
+};
+
+export const Default = () => (
+	<div style={ wrapperStyles }>
+		<crowdsignal-embed src="https://crowdsignal.localhost:9001/B2A9100BBAB921D0?preview=true" />
```
jcheringer, line 19: Since this is using a localhost URL, I couldn't see this correctly when running Storybook.
I changed this URL to one of my projects, and it loaded fine.
Would it be possible to add a production URL here?

jcheringer (changes requested):
Looks great! I loved the `ResizeObserver` + `PostMessage` combo. 👏 
Added a small comment regarding the Storybook URL.
Also, I couldn't make the submit button work. Tried on Storybook and using a local HTML file.
Is this expected? Or maybe I'm doing something wrong?

apps/embed/src/stories/index.js
ice9js: The reason I didn't add a production URL is you wouldn't be able to use it for testing then. And none of this exists on production yet so it wouldn't work yet anyway.

I guess you do need to make sure to run both at the same time: `yarn start` - for the project renderer - and then `yarn storybook`. I haven't had any issues with that myself.

ice9js: Actually, now that I think of it... 🤔  I guess the preview is only accessible to you if you're the project owner. I thought it'd be handy to add the preview there so you can go click through it without submitting unnecessary responses, but in retrospective it maybe wasn't the best idea.  
Have you tried removing `?preview=true`? So you get the public version?

PR #273, opened 2022-06-21
Re-enable the core cover block

ice9js:
This patch re-enables the cover block in our editor that was accidentally removed when we fixed the issue that allowed the user to use any block regardless of the setting in #263.

# Testing

- Create/edit a project.
- You should be able to add cover blocks.
- You should be able to convert image blocks to cover blocks.


jcheringer (approved):
Works for me

PR #274, opened 2022-06-21
use-field hook Refactoring

jcheringer:
## Summary

This is a clean-up/housekeeping PR that aims to simplify the `use-field` hook.
It now doesn't care about the field type. The type-specific logic was moved to the components.
The hook, now, takes a `fieldName` and a flag to define if it can handle multiple values.
It'll return the current `fieldValue` and an `onUpdate` handler to allow values updates.

## Testing Instructions
* Create a project and add at least one of each Crowdsignal block
* Play with the blocks, especially the validations
* Go to the renderer and check if the blocks behave as expected and if the responses are collected correctly

### DIFF
--- a/packages/blocks/src/file-input/index.js
+++ b/packages/blocks/src/file-input/index.js
@@ -13,9 +13,8 @@ import { useColorStyles } from '@crowdsignal/styles';
 import { useField } from '@crowdsignal/form';
 
 const FileInput = ( { attributes, className } ) => {
-	const { inputProps, error } = useField( {
-		name: `q_${ attributes.clientId }_upload`,
-		type: 'file',
+	const { error, fieldValue, onUpdate } = useField( {
+		fieldName: `q_${ attributes.clientId }_upload`,
 		validation: ( files ) => {
 			if ( attributes.mandatory && ( ! files || files.length === 0 ) ) {
 				return __( 'This field is required', 'blocks' );
@@ -60,7 +59,8 @@ const FileInput = ( { attributes, className } ) => {
 			</FormInputWrapper.Label>
 			<FormFileInput
 				attributes={ attributes }
-				inputProps={ inputProps }
+				files={ fieldValue }
+				onChange={ ( event ) => onUpdate( event.target.files ) }
 			/>
 			{ error && <ErrorMessage>{ error }</ErrorMessage> }
 		</FormInputWrapper>
--- a/packages/blocks/src/file-input/index.js
+++ b/packages/blocks/src/file-input/index.js
@@ -13,7 +13,7 @@ import { useColorStyles } from '@crowdsignal/styles';
 import { useField } from '@crowdsignal/form';
 
 const FileInput = ( { attributes, className } ) => {
-	const { error, fieldValue, onUpdate } = useField( {
+	const { error, onChange, fieldValue } = useField( {
 		fieldName: `q_${ attributes.clientId }_upload`,
 		validation: ( files ) => {
 			if ( attributes.mandatory && ( ! files || files.length === 0 ) ) {
@@ -60,7 +60,7 @@ const FileInput = ( { attributes, className } ) => {
 			<FormFileInput
 				attributes={ attributes }
 				files={ fieldValue }
-				onChange={ ( event ) => onUpdate( event.target.files ) }
+				onChange={ onChange }
 			/>
 			{ error && <ErrorMessage>{ error }</ErrorMessage> }
 		</FormInputWrapper>
--- a/packages/blocks/src/matrix-question/cell.js
+++ b/packages/blocks/src/matrix-question/cell.js
@@ -1,4 +1,9 @@
 /**
+ * External dependencies
+ */
+import { filter, uniq } from 'lodash';
+
+/**
  * Internal dependencies
  */
 import { useField } from '@crowdsignal/form';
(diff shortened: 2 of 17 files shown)


packages/form/src/hooks/use-field.js
```diff
@@ -13,83 +13,49 @@ import { STORE_NAME } from '../data';
...
-			setFieldValue( formName, fieldName, initialValue );
-		}
-	}, [] );
-
 	return {
 		error,
-		inputProps,
-		onSort,
+		fieldValue,
+		onUpdate,
```
ice9js, line 59: Nit: should we maybe rename this export to `setFieldValue`?

packages/form/src/hooks/use-field.js
```diff
@@ -13,83 +13,49 @@ import { STORE_NAME } from '../data';
...
-		} else if ( type === 'dropdown' ) {
-			newValue = event;
+	useEffect( () => {
+		if ( initialValue ) {
+			setFieldValue( formName, fieldName, initialValue );
 		}
+	}, [] );
 
-		setFieldValue( formName, fieldName, newValue );
+	const onUpdate = ( value, isSelected ) => {
```
ice9js, line 43: Not a huge fan of the second `isSelected` argument - it seems like we're still trying to do too much here.  
Like I mentioned in the previous comment, we should really only care about the value.

packages/form/src/hooks/use-field.js
```diff
@@ -13,83 +13,49 @@ import { STORE_NAME } from '../data';
...
+	validation,
 } ) => {
 	const { name: formName } = useContext( Form.Context );
 	const { setFieldValue } = useDispatch( STORE_NAME );
 	const { error, validateField } = useValidation( { fieldName, validation } );
-	const { value: currentValue } = useSelect(
-		( select ) => select( STORE_NAME ).getFieldData( formName, fieldName ),
+	const fieldValue = useSelect(
+		( select ) => {
+			const defaultValue = isMultiSelect ? [] : '';
```
ice9js, line 26: If we get rid of `isMultiSelect` you could still pass the default value as a prop when calling the hook.

packages/blocks/src/text-input/index.js
```diff
@@ -49,12 +49,13 @@ const TextInput = ( { attributes, className } ) => {
 				<RawHTML>{ attributes.label }</RawHTML>
 			</FormInputWrapper.Label>
 			<FormTextInput
+				value={ fieldValue }
+				placeholder={ attributes.placeholder }
+				onChange={ ( event ) => onUpdate( event.target.value ) }
```
ice9js, line 54: We should do this inside `<FormTextInput />`, so here you just write `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

packages/blocks/src/rating-scale-answer/index.js
```diff
@@ -42,8 +41,10 @@ const RatingScaleAnswer = ( { attributes, className } ) => {
 		<ButtonAnswer
 			attributes={ attributes }
 			className={ classes }
-			inputProps={ inputProps }
 			isMultiSelect={ false }
+			isSelected={ isSelected }
+			onChange={ ( event ) => onUpdate( event.target.value, isSelected ) }
```
ice9js, line 46: The `event -> value` mapping should be done in our input components. So here it's just `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

packages/blocks/src/ranking-question/index.js
```diff
@@ -38,7 +38,7 @@ const RankingQuestion = ( { attributes, children, className } ) => {
 
 		const [ removed ] = answers.splice( source.index, 1 );
 		answers.splice( destination.index, 0, removed );
-		onSort( answers );
+		onUpdate( map( answers, 'props.attributes.clientId' ) );
```
ice9js, line 41: Note: Since this particular use case is not relying on any inputs - this totally make sense here :)

packages/blocks/src/multiple-choice-answer/index.js
```diff
@@ -55,8 +62,10 @@ const MultipleChoiceAnswer = ( { attributes, className } ) => {
 		<ButtonAnswer
 			attributes={ attributes }
 			className={ classes }
-			inputProps={ inputProps }
 			isMultiSelect={ isMultiSelect }
+			isSelected={ isSelected }
+			onChange={ ( event ) => onUpdate( event.target.value, isSelected ) }
```
ice9js, line 67: The `event -> value` mapping should be done in our input components. So here it's just `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

packages/blocks/src/multiple-choice-answer/index.js
```diff
@@ -46,7 +48,12 @@ const MultipleChoiceAnswer = ( { attributes, className } ) => {
 			<CheckboxAnswer
 				attributes={ attributes }
 				className={ classes }
-				inputProps={ inputProps }
+				isMultiSelect={ isMultiSelect }
+				isSelected={ isSelected }
+				onChange={ ( event ) =>
+					onUpdate( event.target.value, isSelected )
```
ice9js, line 54: The `event -> value` mapping should be done in our input components. So here it's just `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

packages/blocks/src/matrix-question/cell.js
```diff
@@ -10,17 +10,28 @@ import { FormCheckbox } from '../components';
...
+		? fieldValue.includes( column.clientId )
+		: fieldValue === column.clientId;
+
 	return (
 		<MatrixCell as="label">
-			<FormCheckbox { ...inputProps } />
+			<FormCheckbox
+				checked={ isSelected }
+				isMultiSelect={ multipleChoice }
+				onChange={ ( event ) =>
```
ice9js, line 29: The `event -> value` mapping should be done in our input components. So here it's just `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

packages/blocks/src/file-input/index.js
```diff
@@ -60,7 +59,8 @@ const FileInput = ( { attributes, className } ) => {
 			</FormInputWrapper.Label>
 			<FormFileInput
 				attributes={ attributes }
-				inputProps={ inputProps }
+				files={ fieldValue }
+				onChange={ ( event ) => onUpdate( event.target.files ) }
```
ice9js, line 63: The `event -> value` mapping should be done in our input components. So here it's just `onChange={ onUpdate }` or `onChange={ setFieldValue }`.

ice9js (changes requested):
Looks like you might've went a little deep in this refactor - I originally really just wanted to address this part of what's happening in the project renderer:

https://github.com/Automattic/crowdsignal-ui/blob/main/apps/project-renderer/src/components/project-form/use-project-data.js#L63

The idea is our form should return an object on submit that can be turned into `FormData` 1:1, without any additional conversions at that stage. Although I guess what you did is a prerequisite for that as well.

Either way, this is very much appreciated and I do think it will be a nice improvement. I left a few comments regarding getting rid of remaining magic, and moving the mapping logic for `event -> fieldValue` into our form components instead of blocks themselves.  
If you have any counter arguments for why that'd be a bad idea, we can also use a custom `onChangeValue` prop on those form components - so the original `onChange` is still available in case you ever need to access it?

packages/blocks/src/multiple-choice-answer/index.js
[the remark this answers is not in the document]
jcheringer: As you guessed, my idea here was to avoid repetition.
But I agree, maybe it's too much for now.
If it starts to hurt us, we can think about the helpers.

ice9js (approved):
Looks good! Thanks for addressing everything and getting this ready. Feel free to merge after rebase.

PR #275, opened 2022-06-22
Add a card embed format

ice9js:
This patch adds support for a second embed type: project cards. As proposed by @digitalwaveride, the idea is cards would have a fixed 'viewport size' and scale the content to fit the available area on the page while keeping the same aspect ratio.

As they work slightly differently, I figured the cleanest way to implement/share them would be to give them their own component:

```
<crowdsignal-card src="YOUR_PROJECT_URL" viewport-width="1200" viewport-height="600"></crowdsignal-card>
```

Viewport width and height accept pixel values for the desired, let's call it 'preview size' of the project. The card will never exceed these dimensions but it can be smaller, in which case the aspect ratio is preserved and the content is scaled down.

Screenshot 2022-06-22 at 20 21 15

Note that the implementation is still a little quirky and there's still things to figure out, this patch really serves as a proof of concept to get a better feeling for how the format works before committing to more work on it.

# Testing

- Run `yarn start` and `yarn storybook`.
- Inside storybook you should now see a new card embed type.
- Note that with the current example, it's actually impossible to submit it (scrolling is disabled).

### DIFF
--- a/apps/embed/src/card.js
+++ b/apps/embed/src/card.js
@@ -1,0 +1,52 @@
+/**
+ * Crowdsignal embed component.
+ */
+class CrowdsignalCard extends window.HTMLElement {
+	/**
+	 * Reference to the main iframe
+	 */
+	#frame;
+
+	constructor() {
+		super();
+
+		this.attachShadow( { mode: 'open' } );
+	}
+
+	connectedCallback() {
+		const viewportWidth = parseInt(
+			this.getAttribute( 'viewport-width', 10 )
+		);
+		const viewportHeight = parseInt(
+			this.getAttribute( 'viewport-height', 10 )
+		);
+
+		const wrapper = document.createElement( 'div' );
+		wrapper.style.width = '100%';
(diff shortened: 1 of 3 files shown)


apps/embed/src/card.js
```diff
@@ -0,0 +1,52 @@
...
+		const wrapper = document.createElement( 'div' );
+		wrapper.style.width = '100%';
+		wrapper.style.maxWidth = `${ viewportWidth }px`;
+		this.shadowRoot.appendChild( wrapper );
+
+		// Measure the wrapper width after it mounts to determine the scale factor
+		const scale = wrapper.offsetWidth / viewportWidth;
+
+		wrapper.style.height = `${ Math.floor( viewportHeight * scale ) }px`;
+		wrapper.style.height = `${ Math.floor( viewportHeight * scale ) }px`;
```
jcheringer, line 33: `if ( Line32 === Line33 ) addGithubComment()` 😆

PR #276, opened 2022-06-23
Add multiple choice support for the Dropdown block

jcheringer:
## Summary

This PR adds a toolbar option to switch between `Single Choice` and `Multiple Choices` for the Dropdown block.

image

## Testing Instructions

* Create/edit a project and add a Dropdown block
* Check if the Toolbar shows the Multiple/Single choice option
* Go to the Renderer and check if the block works as expected


packages/block-editor/src/dropdown-input/attributes.js
```diff
@@ -24,6 +19,10 @@ export default {
 		type: 'boolean',
 		default: false,
 	},
+	multipleChoice: {
```
CGastrell, line 22: I wonder if we should stick to the Multiple Choice block attributes. We defined there `minimumChoices` and `maximumChoices` which derive into the **choiceType** of the question. Then again, that would force us to raise the level of complexity on this block's Toolbar/Sidebar.

CGastrell (commented):
Anything we can do about this?
image

packages/block-editor/src/dropdown-input/attributes.js
jcheringer: I think this is a 'feature'. When allowing multiple choices, do we want to limit the min/max choices?
If so, indeed, we need to make the sidebar similar to the MC Question.
Otherwise, I'd stick with just a boolean flag and adjust the mapping on the backend, like the Matrix block.
I think @digitalwaveride thoughts will be very appreciated here. 🙂

jcheringer:
@CGastrell I solved the DevTools warning and created a backed diff (D83176-code) to add the `choice_type` mapping.
I think we can proceed with this PR using just a boolean flag for the `choice_type` and, if we decide to limit the min/max choices, we do it in another PR.
Does it sound good?

packages/block-editor/src/dropdown-input/attributes.js
CGastrell: Agree that it's a "feature". When we were working on the MC block, we had this discussion with @ice9js and he made this point: having max/min settings would easily override the need for `multiple` or even `mandatory` (this is something that happens on backend actually):

```
// If mandatory then we always need at least 1 option
if (
	$this->is_mandatory() &&
	$this->min_options == 0
) {
	$this->min_options = 1;
}
...
// If we have a minimum then we also have a mandatory question
if ( $this->min_options > 0 )
	$this->is_mandatory( true );
```
So, as we figured, instead of adding the property "multiple", we added min/max thinking how it would, at some point, make sense for the user to have such option. Reason being, if we added both props/attributes (min/max and multiple), eventually those could collide and a more complex resolving code would be needed. Ex: choose single selection AND minimum choices 2, choose multiple selection and max options 1. 

And, TBH, I think it makes sense and also adds for consistency (as we ARE extending/sharing code from MC block/question).

All this said, we could do this on a revisit, but since we're here...

Design-wise I'm not sure what was the original intention, any opinion/advise @digitalwaveride ?

digitalwaveride: As mentioned in our call today, if we can "steal" the same functionality from the MC block relatively easily, then let's do this. Otherwise, let's postpone that additional feature for now.

jcheringer:
As we discussed, I added the changes to include the `maximumChoices` attribute.
The behavior, now, should be similar to what we have on the MC Question block.
The backend changes are not required anymore.

CGastrell (approved):
Works nicely and, indeed, no backend work needed!

I did find a mishap with the _placeholder_: if I hit enter, I'm allowed to write a second, third... line.
image

Not a "problem" per se, but I noticed 2 things:
- text is centered (no option to change it)
- the renderer shows `<br>`s as carriage return and cuts the text when there's no room for it
image

So, I guess we can go with the "allow multiline" and then fix the styling/alignment and the renderer. Or maybe it's simpler if we, for now, just don't allow this multiline behavior on the placeholder.

Functionally speaking this looks fine to me, but I'll leave it up to you and @digitalwaveride on the final word regarding the scope of those details and if that deserves to be dealt with here or an upcoming PR

jcheringer:
I don't know why it's allowing multi-lines considering this property is set to `false`: https://github.com/Automattic/crowdsignal-ui/blob/main/packages/block-editor/src/dropdown-input/edit.js#L145
Maybe an issue with the RichText component? 🤷

jcheringer:
Decided to take a look at the RichText source code and found out there is a `disableLineBreaks` property.
Tested and it worked. I'm going to push the change.

PR #277, opened 2022-06-27
Dropdown block themes adjustments

jcheringer:
## Summary

This PR updates the Dropdown block style to make it consistent with each theme

## Testing Instructions

* Create/edit a project and add a Dropdown block
* Check if the block style is consistent with each theme on the Editor and the Renderer

### DIFF
--- a/packages/theme-compatibility/src/twentynineteen/base.scss
+++ b/packages/theme-compatibility/src/twentynineteen/base.scss
@@ -64,6 +64,6 @@ button, input, select, optgroup, textarea {
 	}
 }
 
-.crowdsignal-forms-button__button {
+.crowdsignal-forms-button__button, .crowdsignal-forms-dropdown-input__option {
 	font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
 }


jcheringer:
Fixes Automattic/crowdsignal#410

CGastrell:
There's a consistent (throughout all themes) "thing" that feels a bit itchy: selecting the first/deselecting the last option makes the input shorter. It's not a "bad" thing, but it is a little annoying. Selecting/deselecting repeats the effect making the input longer/shorter every time.

Hovering:
image

Selected:
image


Blockbase/Quadrat themes shows a scrollbar on the dropdown (regardless of how many options there are), not shown on other themes. Which makes me wonder, how tall can the dropdown become? Do we set a limit there?

Blockbase:
image

Quadrat:
image

jcheringer:
Regarding the dropdown width, this is an issue I mentioned a few days ago. When using the width `auto`, since the options list is absolutely positioned and `display: none` when closed, I didn't find a way to set the dropdown width considering the largest string in the options.

About the scrollbar, the options wrapper has a `max-height: 210px`. I only see the scrollbar if the number of options makes it bigger than 210px.
On Blockbase, with only two options, I can't see the scrollbar

image

digitalwaveride (changes requested):
Thanks João!!! Looking good!!!

Just one little thingy that I found:

On Twenty Nineteen theme:


Dropdown options are in "serif" font... 
should have though the same font as the "placeholder" text  -> sans serif 
`font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif`

image


image

image

jcheringer:
Nice catch @digitalwaveride!
I've just pushed the changes.

PR #278, opened 2022-06-30
Include P2 Theme

jcheringer:
## Summary

This PR adds the P2 theme as a new theme option for the new Editor.

Depends on: D83496-code

## Test Instructions

* Apply D83496-code
* Create/edit a new project
* Use the available options to change the theme
* You should see the P2 theme as an option
* Select the P2 theme and play around with some blocks to make sure it looks good.

### DIFF
--- a/apps/dashboard/src/components/modal/index.js
+++ b/apps/dashboard/src/components/modal/index.js
@@ -14,7 +14,7 @@ export const ModalWrapper = styled.div`
 	box-sizing: border-box;
 	display: flex;
 	justify-content: center;
-	padding: 100px;
+	padding: 50px;
 	position: fixed;
 	bottom: 0;
 	left: 0;
@@ -36,6 +36,7 @@ export const ModalDialog = styled.div`
 	flex-direction: column;
 	width: 90%;
 	max-width: 1600px;
+	max-height: 100%;
 	padding: 40px 60px;
 `;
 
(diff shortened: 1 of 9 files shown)


jcheringer:
@digitalwaveride I think we'll need a better 'preview' image.
Added the image below just as a placeholder for now.

image

digitalwaveride:
> I think we'll need a better 'preview' image.

Let's take the image from
https://wordpress.com/p2/

image


Which overall reminds me, we should change the preview images of all themes...

digitalwaveride:
Super nice João!!! 
Thanks so much for bringing this over so quickly.

A few notes:

---
- [ ] Overall question blocks:

No grey background color. Please compare with Poll block in P2.

---
- [ ] Dropdown Block:

Chevron Icon needs on the right as much padding as text on the left: 
image

---
- [ ] Rating Block:

The blue border line is a bit too much … could we give it the same grey outline by default as the block itself has?
image

---
- [ ] Input Fields:

Please check field heights... at least 48 if you compare input fields in the contact form block in P2
Also please check the padding for the text inside the input fields. 

---
- [ ] Text Question Block / Template Case:

There seems to be some margin missing below. Just an issue in that specific "customer service feedback" template?
image

---

That's it! 🚀 Thx again João!

jcheringer:
Hey @digitalwaveride! Thanks for taking a look.
I think I addressed all the comments.
Could you give it another spin, please? 🙏

digitalwaveride (approved):
Nice João!
Thank you!!!
🚀

PR #284, opened 2022-07-14
Enable sharing for project embeds and cards

ice9js:
This change updates the sharing dialog to add options for sharing project embeds and 'cards', as well as adding embed card settings in the project sidebar.

Screenshot 2022-07-15 at 08 27 17
Screenshot 2022-07-14 at 14 23 46

Requires D84264-code.

# Testing

- When clicking on 'share' in the editor on a public project, you should now get options for an embed card as well as an embed iframe. Verify that the snippets actually work.
- You will also see new embed card settings in the project sidebar. These are the viewport width and height for the card. Verify that they're saved correctly when you edit them (no autosave). Note that they're not yet used by the embed script itself though.


TheSteveK:
Looks good to me, though I get an error when trying to publish.  I think it's something up with our dev environment.  I'll wait for Thomas' review before approving.

CGastrell (approved):
Works now! :shipit:

PR #285, opened 2022-07-15
Include new Patterns list

jcheringer:
## Summary

This PR adds the necessary changes to (finally 🎉 ) make the new pattern list work.

## Testing Instructions
* Open or create a project
* Click the [ + ] icon to open the Block Library
* Select the Patterns tab
* Check if the new pattern list is available
* Play with the patterns to make sure they are working properly

**Note:** There are some styling glitches with the pattern preview that make the preview taller than necessary. That is not related to these changes, and we'll try to address it in a separate PR.


jcheringer:
Fixes Automattic/crowdsignal#336

TheSteveK (commented):
Question:  Should these patterns be enclosed in a group block?  All of the core patterns are enclosed in a group. Mostly because they all seem to add a background color, but I can think of other advantages of a group, like being able to move the whole pattern around the page or delete it all at once.

Also small nitpick: Ice cream is two words 
Screen Shot 2022-09-19 at 10 42 18 AM

TheSteveK (approved):
We'll get 'em next time.  GO for typo fix!

PR #287, opened 2022-07-20
Update Text Input Form block to support multiline

jcheringer:
## Summary

This PR adds a feature to turn the Text Input Form into a multiline text input if the input height is higher than 70px.

## Testing Instructions
* Open or create a project and add a Text Input Form block
* Increase the input height to a value higher than 70px
* The input now should be a textarea
* Check the Renderer to confirm if it's working as expected

### DIFF
--- a/packages/block-editor/src/text-input/edit.js
+++ b/packages/block-editor/src/text-input/edit.js
@@ -9,12 +9,18 @@ import classnames from 'classnames';
 /**
  * Internal dependencies
  */
-import { FormInputWrapper, FormTextInput } from '@crowdsignal/blocks';
+import {
+	FormInputWrapper,
+	FormTextarea,
+	FormTextInput,
+} from '@crowdsignal/blocks';
 import { useColorStyles } from '@crowdsignal/styles';
 import Sidebar from './sidebar';
 import { useClientId } from '@crowdsignal/hooks';
 
 const EditTextInput = ( props ) => {
+	const MULTILINE_THRESHOLD = 70;
+
 	const { attributes, setAttributes, className, isSelected } = props;
 
 	useClientId( props );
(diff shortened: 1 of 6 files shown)


packages/block-editor/src/text-input/edit.js
```diff
@@ -9,12 +9,18 @@ import classnames from 'classnames';
...
+	FormInputWrapper,
+	FormTextarea,
+	FormTextInput,
+} from '@crowdsignal/blocks';
 import { useColorStyles } from '@crowdsignal/styles';
 import Sidebar from './sidebar';
 import { useClientId } from '@crowdsignal/hooks';
 
 const EditTextInput = ( props ) => {
+	const MULTILINE_THRESHOLD = 70;
```
ice9js, line 22: Nit: generally I'd prefer to place constants outside of any functions.

packages/blocks/src/components/form-textarea/index.js
```diff
@@ -17,8 +17,8 @@ const FormTextarePreviewWrapper = styled.div`
 `;
 
 const Textarea = styled.textarea`
-	border-style: solid;
-	border-width: 1px;
+	//border-style: solid;
```
ice9js, line 20: If we're not using it, let's delete it.

packages/blocks/src/text-input/index.js
```diff
@@ -10,11 +10,18 @@ import { isEmpty } from 'lodash';
...
+	ErrorMessage,
+	FormInputWrapper,
+	FormTextarea,
+	FormTextInput,
+} from '../components';
 import { useField } from '@crowdsignal/form';
 import validator from './validations';
 
 const TextInput = ( { attributes, className } ) => {
+	const MULTILINE_THRESHOLD = 70;
```
ice9js, line 23: Again, I'd place it outside of the function. Better yet, you could later assign it to `TextInput.MULTILINE_THRESHOLD` and use that inside the editor block rather than having two separate definitions for the same thing.

ice9js (changes requested):
Looks good, I think there's just a few minor tweaks we could add to `MULTILINE_THRESHOLD` to make it more manageable in the future.

jcheringer:
> Looks good, I think there's just a few minor tweaks we could add to `MULTILINE_THRESHOLD` to make it more manageable in the future.

Thanks for reviewing it Kuba!
I pushed the changes, could you take a look again, please?

ice9js (approved):
LGTM 👍

PR #288, opened 2022-07-21
Include Navigation Bar and settings on the Editor

jcheringer:
## Summary

This PR adds a few settings which allow showing a navigation bar in the Projects.
Those changes affect only the Editor for now.

Depends on: D84663-code

## Testing Instructions
* Apply backend changes
* Open the Project settings on the right sidebar
* You should see a `Navigation` section
* Changing the toggles should allow you to customize the way you want to see the Navigation Bar


TheSteveK:
Should I be able to see the navigation bar when previewing the project or viewing it via the share URL?

It works fine in the editor.  And the navigation stuff shows up in the metadata.  Did I forget to run a build process on the backend?

jcheringer:
> Should I be able to see the navigation bar when previewing the project or viewing it via the share URL?
> 
> It works fine in the editor. And the navigation stuff shows up in the metadata. Did I forget to run a build process on the backend?

Hey Steve! Thanks for taking a look.
I mentioned it in the PR description. 😬 This PR adds the Navigation Bar only for the Editor. 
I'll handle the Renderer in the next PR.
Also, it's worth mentioning that it still needs some styled adjustments for themes other than Leven.

TheSteveK:
You're right, I totally missed that, sorry :)

TheSteveK (approved):
🚀

digitalwaveride (changes requested):
So nice to play with this!!! Thanks João!!!

A few little  things:

#### Confirmation Page 

The navigation bar should not show on the confirmation page... as "resubmitting" makes not really sense... 

image


#### Dropshadow

The shadow is a bit harsh... could you try something like this:
`drop-shadow(0px 0px 6px rgba(0, 0, 0, 0.02)) drop-shadow(0px 2px 4px rgba(0, 0, 0, 0.08));`

image

#### The Back Link
Two little things:

- Back link should not show on page 1.
- And just a nice to have... not important...  but would it be possible to actually make the link work, and go to the previous page (back)?




Thanks again João!!! Very cool!

jcheringer:
Hey @digitalwaveride, thanks for giving it a spin.

> The navigation bar should not show on the confirmation page... as "resubmitting" makes not really sense...

Completely forgot this one, thanks for catching it.

> The shadow is a bit harsh... could you try something like this:
drop-shadow(0px 0px 6px rgba(0, 0, 0, 0.02)) drop-shadow(0px 2px 4px rgba(0, 0, 0, 0.08));

Sure!

> Back link should not show on page 1.

Could we proceed with the approach of showing the back in a disabled state on the first page?
If we simply don't show the back button on the first page, we could give the user the impression that the "Show back button toggle" isn't working if they are on the first page when playing with the toggle.

> And just a nice to have... not important... but would it be possible to actually make the link work, and go to the previous page (back)?

Yes, that's doable.

CGastrell (approved):
Worked fine on my end! :shipit:

digitalwaveride (approved):
👍

PR #290, opened 2022-07-25
Append ?iframe=1 to iframe embed URLs

ice9js:
This patch adds an `?iframe=1` query param to all iframe embed URLs, to prevent the cookie banner from appearing which isn't necessary inside embeds.

# Testing

- Run `yarn start` and `yarn storybook`
- Inside storybook, validate that both card and iframe embeds have a `?iframe=1` param appended to the urls in the iframe's `src` attribute.
- There should be no cookie banners inside the embeds.

### DIFF
--- a/apps/embed/src/embed.js
+++ b/apps/embed/src/embed.js
@@ -15,7 +15,11 @@ class CrowdsignalEmbed extends window.HTMLElement {
 
 	connectedCallback() {
 		this.#frame = document.createElement( 'iframe' );
-		this.#frame.src = this.getAttribute( 'src' );
+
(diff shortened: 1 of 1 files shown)


CGastrell (approved):
Works as described 👍 

Just a note: I'm not on the GDPR zone, I used a forced proxy to test this. Not sure it might give false positives, just thought it was worth mentioning.

Other than that, code looks good, output looks good :shipit:

ice9js:
>Just a note: I'm not on the GDPR zone, I used a forced proxy to test this. Not sure it might give false positives, just thought it was worth mentioning.

That's taken care of in the implementation of the banner itself though.

PR #291, opened 2022-07-26
Fine tune crowdsignal's embed components

ice9js:
When adding our updated oembed provider to WordPress, I discovered a few issues with our current components:

- Use of shadowRoot inside our components prevents the rendering document from seeing any changes inside the component itself. Normally this isn't an issue, but since Gutenberg renders embed previews in an iframe and is using a similar trick to us, it's preventing the preview from resizing correctly. As we're using iframes anyway, there really isn't any worthwile advantage to 'encapsulating them even more' with shadowRoot.
- If multiple embeds are added on a page through oembed, the script will also be loaded multiple times, and it will try to define our custom elements each time resulting in warnings.

This patch fixes these issues.

# Testing

- You can use `yarn start` in combination with `yarn storybook` to verify that our embeds are no longer rendered inside shadowRoot in developer tools.
- Create a separate html document and try adding our embed script to it multiple times. There shouldn't be any warnings about the same element being defined again.

### DIFF
--- a/apps/embed/src/card.js
+++ b/apps/embed/src/card.js
@@ -12,12 +12,6 @@ class CrowdsignalCard extends window.HTMLElement {
 	 */
 	#frame;
 
-	constructor() {
-		super();
-
-		this.attachShadow( { mode: 'open' } );
-	}
-
 	connectedCallback() {
 		const viewportWidth = parseInt(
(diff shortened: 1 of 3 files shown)


jcheringer (approved):
LGTM!

PR #294, opened 2022-07-27
Include the Navigation Bar on the Project Renderer

jcheringer:
## Summary

This PR is a complementary PR to #288.
It includes the Navigation Bar on the Project Renderer.

Depends on: D84663-code

## Testing Instructions

* Apply backend changes
* Open/Create a project
* Enable the settings to show the Navigation Bar
* Publish the project
* Go to the Project Render
* Check if the Navigation Bar works as expected

### DIFF
--- a/packages/form/src/components/form/index.js
+++ b/packages/form/src/components/form/index.js
@@ -20,7 +20,7 @@ const Form = forwardRef( ( { children, name, onSubmit, ...props }, ref ) => {
 		[ name ]
 	);
 
-	const { startSubmit, stopSubmit, initForm } = useDispatch( STORE_NAME );
+	const { startSubmit, stopSubmit } = useDispatch( STORE_NAME );
 
 	const registerValidation = ( fieldName, validation ) =>
 		( validations[ fieldName ] = validation );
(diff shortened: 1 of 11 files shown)


jcheringer:
> It works as expected, but it's invisible to screen readers. I think if you add some Aria code to the container div like Region https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role and then add aria-label to each of the progress dashes that should do the trick.
> 
> I'll leave it to you to decide if you want to tackle that in this PR or if it should be handled later.

Well, I love the idea of addressing the accessibility issues. I think this should be a bigger compromise, though.
So far, we didn't pay much attention to accessibility stuff on the new Editor, maybe we could turn this into a project.

TheSteveK:
Yeah we can pick it up later. It will probably be a thing though TBH :) 

But you are clear to launch 🚀

TheSteveK (approved):
LGTM 🚀

packages/form/src/components/form/index.js
```diff
@@ -40,9 +40,7 @@ const Form = forwardRef( ( { children, name, onSubmit, ...props }, ref ) => {
 
 		startSubmit( name );
 
-		onSubmit( data )
-			.then( () => initForm( name, {} ) )
-			.finally( () => stopSubmit( name ) );
+		onSubmit( data ).finally( () => stopSubmit( name ) );
```
ice9js, line 43: Are we sure this won't cause the data for the previous page to submit when the next page is submitted?
For example, submit page 1, go to page 2, the request for page 2 will contain answers for both page 1 and page 2.

I'd recommend to give each page an appropriate suffix to the name, so we just select the correct set depending on what `currentPage` is.

jcheringer: Thanks for catching that!
Already made the change to add a suffix to the form name based on the page index.

ice9js (approved):
👍

PR #297, opened 2022-08-02
Prevent external styles form affecting embed components

ice9js:
This patch resolves some of the issues noticed by @digitalwaveride  when using our embeds on WordPress' P2 theme.  
While the issues come from the theme itself and affect all embeds, there are still a few things we can do:

In #291, I removed the use of `shadowRoot` from our web components however it seems the P2 theme (and presumably other themes/websites out there) contains CSS styles which target the `iframe` inside our web component directly messing up the layout.

To fix this, I reintroduced `shadowRoot` and added explicit `height` and `width` attributes on the web component itself, so these can be tracked/picked up by a script correctly if our embed is rendered within another iframe - like it is in the Gutenberg editor.

# Testing

- Run `NODE_ENV=production yarn workspace @crowdsignal/embed build`.
- Replace `app.crowdsignal.com/ui/stable/embed/main.js` with the newly built `apps/embed/dist/main.js`.
- Create a new post on a WP.com site with the P2 theme enabled.
- Paste a link for both a regular and a card embed (append `?type=card` for a card embed)
- Publish/preview the post and ensure the embeds are sized correctly.

### DIFF
--- a/apps/embed/src/card.js
+++ b/apps/embed/src/card.js
@@ -12,6 +12,12 @@ class CrowdsignalCard extends window.HTMLElement {
 	 */
 	#frame;
 
+	constructor() {
+		super();
+
+		this.attachShadow( { mode: 'open' } );
+	}
+
 	connectedCallback() {
 		const viewportWidth = parseInt(
 			this.getAttribute( 'viewport-width', 10 )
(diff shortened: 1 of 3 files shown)


jcheringer (approved):
The code looks good to me.
I'm not sure if I tested it correctly, though.
It seemed good, but I'd suggest that someone else double check

PR #300, opened 2022-08-03
Update Share Modal embed buttons

jcheringer:
## Summary

This PR adds two more buttons to the embed sharing options to allow the user to copy the project's embed link.

Design: yKrlfbUJTFt0EMKheKh15c-fi-1953%3A37219

Fixes Automattic/crowdsignal#439

## Testing Instruction
* Open or create a project
* Make sure the project is published
* Click the Share button
* Check if the embed options match the Figma design and work correctly

### DIFF
--- a/apps/dashboard/src/components/share-modal/share-card/share-card-button.js
+++ b/apps/dashboard/src/components/share-modal/share-card/share-card-button.js
@@ -1,0 +1,85 @@
+/**
+ * External dependencies
+ */
+import { useState } from '@wordpress/element';
+import classnames from 'classnames';
+import styled from '@emotion/styled';
+
+/**
+ * Internal dependencies
+ */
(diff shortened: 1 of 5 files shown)


TheSteveK (approved):
Looks good, code copies as expected.

A side note: the modal itself is not accessible.  This was the case before your change, so probably not something to resolve here.  But, when using a keyboard or screen reader, the modal never grabs focus so when you navigate or tab from the share button it just starts going through the elements in the editor behind the modal.

digitalwaveride (approved):
Thanks João for bring this in so quickly! Looking good, and I think we can merge it.

One thing though, not related to that PR directly, the modal window seems not to be responsive at all.
The "cards on the modal" should stack!

I will create a separate issue for that:

image

image

PR #304, opened 2022-08-05
Fix Templates Order

jcheringer:
## Summary

As per @digitalwaveride recommendation, we need to update the order in which the templates are shown.
The recommended order is:

1. Empty Canvas
2. Customer Service Feedback
3. Team Member Onboarding Survey
4. Product Feedback
5. Job Listing – Application Form
6. Event Registration Form
7. Customer Satisfaction Survey
8. Join Waitlist
9. Join Our Community

Also, as the template list was being exported as a JS module, it seems that the templates inside the module were always in alphabetical order when building the bundle for production.
To ensure we get the order we want, I needed to export the templates as an array explicitly.

## Testing Instructions

* Open the project page and check if the templates are in the correct order
* Run `yarn build`
* Copy all the content inside the `dist/` folder and paste inside the folders `stable/` and `next/` on `app.crowdsignal.com/public_html/ui/` from your Sandbox
* Make sure you're sanboxed and open https://app.crowdsignal.com/project
* Check if the templates are in the correct order

### DIFF
--- a/apps/dashboard/src/components/editor/index.js
+++ b/apps/dashboard/src/components/editor/index.js
@@ -9,7 +9,7 @@ import { useEffect } from '@wordpress/element';
  * Internal dependencies
  */
 import HeaderMeta from '../header-meta';
-import { blankProjectTemplate } from '../new-project-wizard/templates';
+import { blankProjectTemplate } from '../new-project-wizard/templates/blank';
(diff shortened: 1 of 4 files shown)


CGastrell (approved):
LGTM! Ordered as described

PR #305, opened 2022-08-11
Fix Previews block width issue

jcheringer:


TheSteveK (approved):
It works as described, very subtle in a tiny window :)  I noticed you updated one template's column blocks to full width from wide.  Do you need to do that to all of them? the satisfaction survey, for example, has a wide column, but maybe that's what it's supposed to be.  

🚀

jcheringer:
> It works as described, very subtle in a tiny window :) I noticed you updated one template's column blocks to full width from wide. Do you need to do that to all of them? the satisfaction survey, for example, has a wide column, but maybe that's what it's supposed to be.
> 
> 🚀

Yeah, the thing is that the previews are simulating a 1200px wide screen, so when you're switching from wide (max-width 1080px) to full (width limit), there isn't much more space to grow.
In the template previews, I changed it to simulate a 1380px screen, so the difference is more noticeable.
Maybe we should change this width on the page navigation too.
About changing the width to full on the templates, I changed one just to highlight the difference so that @digitalwaveride can compare.
I can update a few more if we think it's better.

PR #306, opened 2022-08-19
Include Web Popup sharing option

jcheringer:
## Summary

This PR includes a new sharing option: The Web Popup.

image

## Testing Instructions
* Make sure the project is published
* Click the Share button
* You should see the Popup option
* Copy the WP link, paste it on a WP site and check if the popup is shown correctly.
* Copy the code snipped, paste into an HTML file and open it up on your preferred browser. Check if the popup appears correctly

### DIFF
--- a/apps/embed/src/popup.js
+++ b/apps/embed/src/popup.js
@@ -1,0 +1,174 @@
+/**
+ * Crowdsignal embed component.
+ */
+class CrowdsignalPopup extends window.HTMLElement {
+	/**
+	 * Wrapper container reference
+	 */
+	#wrapper;
+
+	/**
+	 * Close Button reference
+	 */
+	#closeButton;
+
+	/**
+	 * Reference to the main iframe
+	 */
+	#frame;
+
+	connectedCallback() {
+		const isWpEditor =
+			this.ownerDocument.body.className.indexOf( 'wp-block-embed' ) !==
+			-1;
+
+		if ( isWpEditor ) {
+			this.#showPlaceholder();
+		} else {
+			this.#showPopup();
+		}
+
+		window.addEventListener( 'message', ( event ) => {
+			if (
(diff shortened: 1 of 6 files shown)


apps/embed/src/popup.js
```diff
@@ -0,0 +1,174 @@
...
+	 */
+	#closeButton;
+
+	/**
+	 * Reference to the main iframe
+	 */
+	#frame;
+
+	connectedCallback() {
+		const isWpEditor =
```
jcheringer, line 21: This might not be the best way to check if it's inside the editor, but I didn't find a more elegant way to do that.

ice9js (approved):
Overall, looks alright to me.

I'd argue that all the embed should really do is provide the button by itself, inline, and it's up to the client - in this case a WP site - to make sure the code/block goes into the right place.
This also does away with any confusion regarding content appearing in unexpected places and having to explain that to the user.

That's the way I'd do it but if the decision has been made then I guess this is the next best thing.

jcheringer:
> 

I think I didn't fully understand what you meant.
Are you suggesting we should leave it up to the client to decide where to place the popup?

ice9js:
>Are you suggesting we should leave it up to the client to decide where to place the popup?

Yeah, although I was thinking in lines of a 'button-popup' here. So one where you'd have a trigger as well - which is always visible anyway.  
In case of a regular popup embed,... I wish there was a more general way of doing things but offering a preview within the embed itself is really a novelty so I guess it's a start.

jcheringer:
> > Are you suggesting we should leave it up to the client to decide where to place the popup?
> 
> Yeah, although I was thinking in lines of a 'button-popup' here. So one where you'd have a trigger as well - which is always visible anyway. In case of a regular popup embed,... I wish there was a more general way of doing things but offering a preview within the embed itself is really a novelty so I guess it's a start.

I see. Yeah, that should be just a v1.
We can look for ways to improve it in the future.

PR #308, opened 2022-08-23
add permalink handling to editor

TheSteveK:
Note, needs D86379 for API changes both must be applied to sandbox for testing.

Overview:
This change brings our project section of the editor sidebar closer in alignment with the native WordPress.com post/page editor. The Status & Visibility and permalink sections were combined into a single section called Summary.

Screen Shot 2022-08-25 at 1 25 48 PM

A slight update was made to Visibility to add a header with a close button to match WordPress.com and the same header was added to the new URL section.

The URL section is the main attraction, of course.  It features the new ability for the user to update the project's slug.  

Screen Shot 2022-08-25 at 1 27 44 PM

In addition, the API was updated to both allow for slug editing as well as a new behavior.  The backend will only change the slug to the tile in the event there is no slug set already.  Once the slug is set (whether it's by adding a title, or manually setting a slug) the back end will not attempt to change the slug unless a new slug is set by the user.

Testing instructions:
Be sure to also apply D86379 for the necessary API changes.

Create a new project.  The URL button in the project sidebar will be disabled until the project can be published (a first draft is saved).  Save a first draft, but do not change the title.

Ensure you can click on the URL to open the permalink dialogue.  If a title was not set, you will see the permalink field is blank.  Don't make any changes to the permalink and update the project's title and save the project.  Check the permalink and verify it's set to the title now.

Change the title and save again, verify the permalink has not changed.

Now change the permalink.  Notice the message changes under the text field to let the user know they need to save the project to change the permalink.  Save the project and then verify it indeed has the new permalink.


digitalwaveride:
If I try to clock on the "URL" field to edit the URL, I am seeing this:

image



The copied error:
```
Error: A cross-origin error was thrown. React doesn't have access to the actual error object in development. See https://reactjs.org/link/crossorigin-error for more information.
    at Object.invokeGuardedCallbackDev (https://app.crowdsignal.com/js/vendor/react-dom.development.js:4002:21)
    at invokeGuardedCallback (https://app.crowdsignal.com/js/vendor/react-dom.development.js:4053:33)
    at beginWork$1 (https://app.crowdsignal.com/js/vendor/react-dom.development.js:23994:9)
    at performUnitOfWork (https://app.crowdsignal.com/js/vendor/react-dom.development.js:22809:14)
    at workLoopSync (https://app.crowdsignal.com/js/vendor/react-dom.development.js:22737:7)
    at renderRootSync (https://app.crowdsignal.com/js/vendor/react-dom.development.js:22700:9)
    at performSyncWorkOnRoot (https://app.crowdsignal.com/js/vendor/react-dom.development.js:22323:20)
    at https://app.crowdsignal.com/js/vendor/react-dom.development.js:11357:28
    at unstable_runWithPriority (https://app.crowdsignal.com/js/vendor/react.development.js:2764:14)
    at runWithPriority$1 (https://app.crowdsignal.com/js/vendor/react-dom.development.js:11306:12)
```

but maybe just my machine?

CGastrell:
Note: Sidebar closes once the first draft is saved. Annoying.

As soon as I open the popup, the close button has the focus and its popover hint shows:
image
Any way to remove the focus from it? Maybe set it to the text input?

Also on that screenshot, the link at the bottom of the popup is vertically aligned with the link icon on the right, is that what we want?

It would be nice if hitting Enter/Return would set title and close the popup.

Emptying the input field, closing and saving does nothing. Opening the popup again shows the input empty but the slug remains.

This works, I had no issues getting it up and running and it does the job. Amazed actually: I seem to remember having other projects titled `untitled-project`, which I set while testing this one.

Worth noting: 
- set my project title to "untitled project", ended up being "untitled-project" 👍 
- changed title to "some project", turned to "some-project" 👍 
- auto save doesn't trigger 🤔 
- save draft, back to "published" state, project is now viewable at `/some-project` 👍 
- project is also viewable at its old title `/untitled-project` 🤔 
- repeated the process a couple of times, now project has 4 or 5 slugs, all pointing at the same project 🤔 

Wondering what's going on with the slugs and my expectation was that if I changed the slug, the old one wouldn't be accessible anymore.

CGastrell:
Another thing regarding UX: the close button "X" is usually for *canceling*. There should be a way to cancel without modifying anything. Maybe have another button/link there with "Set" or "Set slug/url". 
Once the popup is open and the slightest change made to the input, there's no turning back.

digitalwaveride:
In general, let's keep things simple here, and let's stay with the interaction behavior as on WP.com - besides that additional notification that the user needs to update before the slug change gets live.

> Wondering what's going on with the slugs and my expectation was that if I changed the slug, the old one wouldn't be accessible anymore.

Same on WP.com. Let's keep that.

> Any way to remove the focus from it? Maybe set it to the text input?
+1

digitalwaveride:
In general, let's keep things simple here, and let's stay with the interaction behavior as on WP.com - besides that additional notification that the user needs to update before the slug change gets live.

> Wondering what's going on with the slugs and my expectation was that if I changed the slug, the old one wouldn't be accessible anymore.

Same on WP.com. Let's keep that.

> Any way to remove the focus from it? Maybe set it to the text input?
+1

TheSteveK:
Per Thomas' comment, the only thing left was to fix focus on the close button.  That appeared to be something built into the PopoverHeader component I was using.  It's also experimental (released about 90 days ago) I decided to swap that out for an old-fashioned div, H2 and our own modal close component.

DustinHartzler:
Tested and worked as expected.

CGastrell (approved):
Let's get this in, we'll give it another spin later :shipit:

PR #317, opened 2022-09-08
Move editor settings to the store

jcheringer:
## Summary

This PR changes the way we handle the editor settings.
We're moving the settings to the store to keep it synced between the Crowdsignal editor and the Isolated Block editor.
This should solve a few styling issues, especially on previews, caused by the Isolated Block Editor overriding some injected styles.

The blocks previews and the patterns inside the block library should look better
image

image

The pages previews should be more consistent with the editor's content
image

**Note**: On the development environment, there's still a styling issue where the theme's primary color is being overridden, making some elements, like buttons, have a black color instead of the theme color. We'll try to fix this issue on another PR.
On the production build, it should look good.

## Testing Instructions
* Checkout the branch, build the project and deploy it to your sandbox
* Make sure you're sandboxed and open/create a project using the production's URL
* Look at the styles mentioned above and check if they look good.


apps/dashboard/src/components/editor/editor.js
```diff
@@ -108,23 +108,34 @@ const Editor = ( { project } ) => {
 
 	const settings = useMemo( () => {
 		if ( ! confirmationPage ) {
-			return baseSettings;
+			return editorSettings;
 		}
 
-		return tap( cloneDeep( baseSettings ), ( { editor, iso } ) => {
+		return tap( cloneDeep( editorSettings ), ( { editor, iso } ) => {
```
ice9js, line 114: I wonder, should we move this to useEditorContent, or can we even maybe incorporate it into `getEditorSettings`?
It's the same store so you do have access to `confirmationPage` there, the trick would be to not create a new object every time that selector is run as that may cause unnecessary reloads.

You'd need some sort of memoization to solve that - I think `reselect` is the most straight forward approach historically, and we've actually used it within the old frontend too so I think it's a safe one to install.

ice9js (changes requested):
Nice! This is quite an improvement in making the settings code much more straightforward and more reliable, thanks Joao!

I think we could clean it up even more and remove the settings `useEffect` from `editor.js` altogether and incorporate it right into the `getEditorSettings()` selector, we might as well do it while we're at it.

Other than that I noticed that we still might need to look into at some point, the colors on the previews don't always match what's in the editor. This is an issue for pretty much all our blocks, but I wouldn't consider it a blocker as this is already a big improvement over the previous version where inserter previews didn't work at all. Just something we should work on later. Attaching screenshots to highlight what I mean:

Screenshot 2022-09-09 at 12 16 30
Screenshot 2022-09-09 at 12 16 39

jcheringer:
> Other than that I noticed that we still might need to look into at some point, the colors on the previews don't always match what's in the editor. This is an issue for pretty much all our blocks, but I wouldn't consider it a blocker as this is already a big improvement over the previous version where inserter previews didn't work at all. Just something we should work on later. Attaching screenshots to highlight what I mean:

Yeah, I noticed this issue as well. That's what I tried to explain on the note I put on the description above, but I might not have been clear enough 😅 
But I only faced this issue in the development environment. If you publish the changes to your sandbox, it should work properly.
I think I know what's causing the problem, but I don't know yet the best way to solve it. 
I want to discuss it with you in another thread.

apps/dashboard/src/components/editor/editor.js
jcheringer: Thanks for the suggestion and for pointing out the `reselect` lib.
I tried to achieve what you suggested with [e8d7994](https://github.com/Automattic/crowdsignal-ui/pull/317/commits/e8d799460c3381c158b1767b993f05af29bde407).
Let me know if you thought something different or if something is missing.

apps/dashboard/src/data/editor/selectors.js
```diff
@@ -2,7 +2,26 @@
...
+	some,
+	tap,
+} from 'lodash';
+import { createSelectorCreator, defaultMemoize } from 'reselect';
+
+// Creates a 'createSelector' instance using lodash 'isEqual'
+// function as the comparison method
+const createDeepEqualSelector = createSelectorCreator(
+	defaultMemoize,
+	isEqual
```
ice9js, line 23: Are you sure you need deep equality here? The objects are nested, yes, but since everything comes from Redux stores, we should be able to rely on regular `createSelector`.

ice9js (changes requested):
Nice, I left just one more comment - I think we can just use `createSelector` with no need for deep comparisons for `getEditorSettings()`.

apps/dashboard/src/data/editor/selectors.js
jcheringer: As we discussed, I removed the deep equality here and changed the styles-resolver logic a bit to prevent unnecessary settings updates.

ice9js (approved):
Nice, let's merge this 👍

PR #319, opened 2022-09-21
add datepicker block

TheSteveK:
This PR adds a new date picker block, using the React-Datepicker https://github.com/Hacker0x01/react-datepicker package.  There is a native Gutenberg component, however it would have required heavier dependencies, so this date picker was chosen as it is more light weight. 

Screen Shot 2022-09-23 at 2 59 29 PM

Unlike the blocked used for Jetpack forms, this block displays the selected (or current initially) date to help the users understand what to do with the box. 

Testing:
Install this patch and the backend code.  You will need to re-run yarn install to install the new dependency

Fire up a test project and add the date picker block.  Play around with it, make sure it makes sense.

Publish the project and then open the preview and remove the ?preview=true from the query string and reload.  Select and submit your date along with any other answers.

Check the results for the project and verify the date has been registered along with any other data.

Requires D88396


packages/block-editor/src/date-time-picker/edit.js
```diff
@@ -0,0 +1,59 @@
...
+					placeholder={ __(
+						'Enter date picker label',
+						'block-editor'
+					) }
+					onChange={ handleChangeLabel }
+					value={ attributes.label }
+				/>
+			</FormInputWrapper.Label>
+			<DateTimePicker
+				className={ classes }
```
jcheringer, line 49: I guess those classes should be on the root element.

packages/block-editor/src/date-time-picker/edit.js
```diff
@@ -0,0 +1,59 @@
...
+				selected={ startDate }
+				onChange={ ( date ) => setStartDate( date ) }
```
jcheringer, line 51: I'm not sure if this was discussed with Thomas, but usually, on the Editor, we don't need to implement the block public behavior.
In this case, I'd say that having the current date as the default value for the input and showing the calendar when the block is selected should be enough for the user to preview the block. I think the `onChange` isn't necessary here.

packages/block-editor/src/date-time-picker/attributes.js
```diff
@@ -0,0 +1,32 @@
...
+		default: '',
+	},
+	date: {
+		type: 'string',
+		default: null,
+	},
+	time: {
+		type: 'string',
+		default: null,
+	},
```
jcheringer, line 17: I guess you don't need those two attributes, right?
On the Renderer version, the selected values can be stored in the component's local state and sent to the backend when the page is submitted.

packages/blocks/src/date-time-picker/index.js
```diff
@@ -0,0 +1,60 @@
...
+			style={ { ...useColorStyles( attributes ) } }
+		>
+			<FormInputWrapper.Label className="crowdsignal-forms-text-input-block__label">
+				<RawHTML>{ attributes.label }</RawHTML>
+			</FormInputWrapper.Label>
+			<DatePicker
+				className={ classes }
+				selected={ startDate }
+				dateFormat="MMMM d, yyyy"
+				onChange={ ( date ) => setStartDate( date ) }
```
jcheringer, line 51: Here you probably want to use the `fieldValue` property. 
This will link your block with the Redux store.

Then, to update the store value, you need to call the `onChange` handler exported by the `useField` hook.

```JavaScript
<DatePicker
	...
	selected={ fieldValue }
	onChange={ onChange }
	...
/>
```

If you need to initialize the block with some value, the `useField` hook accepts an `initialValue` property.
And, if you need to parse the selected date somehow, create a handler instead of passing the `onChange` directly to the `<DatePicket />` component.

jcheringer (commented):
Added a few comments, but since this is a draft, please ignore them if they are already on your TODO list. 🙂

packages/block-editor/src/date-time-picker/edit.js
```diff
@@ -0,0 +1,61 @@
...
+import Sidebar from './sidebar';
+import { useClientId } from '@crowdsignal/hooks';
+
+const EditDateTimePicker = ( props ) => {
+	const { attributes, setAttributes, className } = props;
+
+	useClientId( props );
+
+	const handleChangeLabel = ( label ) => setAttributes( { label } );
+	const classes = classnames(
```
DustinHartzler, line 24: Small nitpick, but all the others have a space between the lines, probably should stay consistent.

packages/block-editor/src/date-time-picker/edit.js
```diff
@@ -0,0 +1,61 @@
...
+				selected={ new Date() }
+				popperModifiers={ [
+					{
+						name: 'arrow',
+						options: {
+							padding: 100,
+						},
+					},
+				] }
+			/>
```
DustinHartzler, line 56: Also by using this `<DatePicker>` component, you don't get the required field icon on the editor side like the rest of our blocks.

packages/block-editor/src/date-time-picker/edit.js
TheSteveK: ✔️ Thanks!  good catch

packages/block-editor/src/date-time-picker/edit.js
TheSteveK: 🤔  ok, going to have to work that one out, must have to do with my using the raw component

packages/block-editor/src/date-time-picker/edit.js
```diff
@@ -0,0 +1,64 @@
...
+		}
+	);
+
+	return (
+		<FormInputWrapper
+			style={ { ...useColorStyles( attributes ) } }
+			className={ classes }
+		>
+			<Sidebar { ...props } />
+			<FormInputWrapper.Label className="crowdsignal-forms-text-input-block__label">
```
jcheringer, line 39: Let's change this to match the block name.

packages/blocks/src/date-time-picker/index.js
```diff
@@ -0,0 +1,74 @@
...
+			'is-error': error,
+		}
+	);
+
+	return (
+		<FormInputWrapper
+			className={ classes }
+			style={ { ...useColorStyles( attributes ) } }
+		>
+			<FormInputWrapper.Label className="crowdsignal-forms-text-input-block__label">
```
jcheringer, line 50: Same here about matching the block name

jcheringer (changes requested):
Added two small comments about the block CSS class.

packages/block-editor/src/date-time-picker/edit.js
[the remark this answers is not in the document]
jcheringer: I guess It might be worth showing the popover on the editor to give the users a hint on how It'll look on the renderer.
I think we don't need to allow date changes, though.

jcheringer:
On Twenty-Twenty-Two (both) and P2 theme, the styles are not looking very good.

image

packages/block-editor/src/date-time-picker/styles.js
```diff
@@ -0,0 +1,19 @@
+/**
```
jcheringer, line 1: These styles are the same as the styles in the `blocks` package.
We can avoid this duplication by creating a `react-datepicker` extension inside `blocks/src/components`.

Following the naming we already have there you'd create something like: `packages/blocks/src/components/form-date-picker/index.js`

```JavaScript
/**
 * External dependencies
 */
import DatePicker from 'react-datepicker';
import styled from '@emotion/styled';

//Feel free to move these styles to another file if you want
export const DatePickerWrapper = styled.div`
	input,
	textarea {
		color: var( --wp--preset--color--foreground );
		background-color: transparent;
		border-width: 1px;
		border-style: solid;
		padding: 8px;
	}

	.react-datepicker__day--selected {
		color: var( --wp--preset--color--background );
		background-color: var( --wp--preset--color--primary );
	}
`;

const FormDatePicker = ( props ) => {
	return (
		<DatePickerWrapper>
			<DatePicker
				dateFormat="MMMM d, yyyy"
				popperModifiers={ [
					{
						name: 'arrow',
						options: {
							padding: 100,
						},
					},
				] }
				{ ...props }
			/>
		</DatePickerWrapper>
	)
}

export default FormDatePicker;
```

That would allow you to also add some default values to properties like `dateFormat` and `popperModifiers`.
This way, you could use the `<FormDatePicker />` on both packages, `blocks` and `block-editor`

On `blocks/src/date-time-picker/index.js`:
```JavaScript
		<FormInputWrapper
			className={ classes }
			style={ { ...useColorStyles( attributes ) } }
		>
			<FormInputWrapper.Label className="crowdsignal-forms-date-time-picker-block__label">
				<RawHTML>{ attributes.label }</RawHTML>
			</FormInputWrapper.Label>
			<FormDatePicker
				selected={ parsedDate }
				onChange={ handleDateChange }
			/>
			{ error && <ErrorMessage>{ error }</ErrorMessage> }
		</FormInputWrapper>
```

On `block-editor/src/date-time-picker/edit.js`:
```JavaScript
		<FormInputWrapper
			style={ { ...useColorStyles( attributes ) } }
			className={ classes }
		>
			<Sidebar { ...props } />
			<FormInputWrapper.Label className="crowdsignal-forms-date-time-picker-block__label">
				<RichText
					placeholder={ __(
						'Enter date picker label',
						'block-editor'
					) }
					onChange={ handleChangeLabel }
					value={ attributes.label }
				/>
			</FormInputWrapper.Label>
			<FormDatePicker
				selected={ new Date() }
			/>
		</FormInputWrapper>
```

jcheringer (changes requested):
Adding a comment explaining how we could make things a bit cleaner and avoid the style duplication.
Let me know if I can clarify anything 🙂

packages/block-editor/src/date-time-picker/styles.js
TheSteveK: Good stuff, I just pushed a commit with your suggestions.  I'll tackle the other CSS next

jcheringer (approved):
LGTM! 🚀

PR #320, opened 2022-09-22
Include new Editor Previews

jcheringer:
## Summary

This PR changes the Preview button to be more similar to the WP Preview button.
By clicking the button, it'll open a popover with some preview options.

image

The `Desktop`, ` Tablet`, and `Mobile` options were already on the Isolated Block Editor. We're just exposing them here.

The `Popup` and `Embed Card` are Crowdsignal specific and should simulate those distribution channels.
They will use the defined embed settings to ensure the preview is as close as possible to the public version.

> There's a known preview styling issue that makes some elements have a black color instead of the theme color. 
That's not caused by this PR and affects only the local environment. On production, it works fine.

## Testing Instructions

* Run `yarn install & yarn start`
* Open/create a project and test the new preview options to ensure they are working correctly.


DustinHartzler:
Functionality works as expected.

For the Quadrat theme and Blockbase themes, the non-Desktop previews styling is a little wonky.

Here's Quadrat for example:
Quadrat

TheSteveK (approved):
Works for me!

PR #321, opened 2022-09-26
Fix Preview Styles

jcheringer:
## Summary

Our page previews are not looking very good, especially for themes other than Leven.
It's due to a [Gutenberg logic](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/iframe/index.js#L360-L363) that tries to get styles from the parent document and include them into the previews.
This logic fails to get external styles, and since it's adding some `<link>` tags inside the iframe body, those styles are overriding the styles from the `<head>` and messing stuff.

This PR includes a script that will try to remove these `<link>` tags from the iframe body.

## Testing Instructions

* Open/create a project
* Switch to a few themes or all of them and check if the previews look better

### DIFF
--- a/apps/dashboard/src/components/editor/styles-resolver.js
+++ b/apps/dashboard/src/components/editor/styles-resolver.js
@@ -16,6 +16,10 @@ window.__editorAssets = window.__editorAssets || {
 	scripts: '',
 };
 
+const IFRAME_SELECTOR =
+	'.block-editor-block-preview__container iframe, .edit-post-visual-editor__content-area iframe';
+const WRAPPER_DIV_SELECTOR = 'body > div:first-child';
+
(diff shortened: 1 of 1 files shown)


TheSteveK:
It's better, but there's still some spacing difference between the title and the blocks when the heading is to the left.  Is that ok?

Editor:

Screen Shot 2022-09-26 at 10 40 06 AM

Preview:

 
Screen Shot 2022-09-26 at 10 41 30 AM

jcheringer:
Ohh, I guess I wasn't clear enough about what this fixes @TheSteveK, sorry!
This PR changes stuff for the previews inside the Editor (Page Previews, Patterns, Blocks previews).
It changes nothing on the Renderer.

TheSteveK (approved):
Oh, then in that case, looks like it works :)

PR #322, opened 2022-09-28
New/phone number block

DustinHartzler:
Created a new block for Phone Number, complete with a drop down country icon selector.

To test:

- Apply D88654 backend changes
- Add a phone number block
- Test the settings (default country code, whether the countries appear or not)


packages/blocks/src/phone-number/index.js
```diff
@@ -0,0 +1,76 @@
...
+				<Input
+					defaultCountry={ attributes.country.toUpperCase() }
+					onChange={ onChange }
+					placeholder={ attributes.placeholder }
+					style={ {
+						width: attributes.inputWidth,
+					} }
+					value={ fieldValue }
+				/>
+			) }
```
DustinHartzler, line 68: Is there a more elegant way to write this? the `<PhoneInput>` is the version with the flag and `<Input>` is the non-flagged version.

packages/block-editor/src/phone-number/edit.js
```diff
@@ -0,0 +1,99 @@
...
+						defaultCountry={ attributes.country.toUpperCase() }
+						placeholder={ __(
+							'Enter Phone Number',
+							'block-editor'
+						) }
+						value={ attributes.placeholder }
+						onChange={ handleChangeCountry }
+						className="crowdsignal-forms-phone-number-block__wrapper"
+					/>
+				) }
```
DustinHartzler, line 81: This works to display the flag on the editor when the toggle switch is enabled, but I don't want it to be able to be an active drop down.

Did I overlook an option here?
https://www.npmjs.com/package/react-phone-number-input

jcheringer: It seems there's no way to show the Country Flag dropdown in a disabled state. 
Adding the CSS `pointer-events: none` would do the job, but I'd suggest leaving the behavior as is and just avoid handling the country change by passing the `noop` function from lodash to the `onChange` event:

```JavaScript
					<PhoneInput
						...
						onChange={ noop }
						...
					/>
```

packages/blocks/src/phone-number/index.js
jcheringer: Yes. You could move this piece of code to a new React component. 
A short way would be something like this:

```JavaScript
	const PhoneBlock = ( props ) => ( flag ? <PhoneInput { ...props } /> : <Input { ...props } /> );
	
	//Now you could do this:
	<PhoneBlock
		defaultCountry={ attributes.country.toUpperCase() }
		placeholder={ __(
			'Enter Phone Number',
			'block-editor'
		) }
		value={ attributes.placeholder }
		onChange={ noop }
		className="crowdsignal-forms-phone-number-block__wrapper"
	/>
```
Here I'm getting the `flag` value from the outer context, the main component. 
But you could accept the flag as a prop and pass it when instantiating the component.

However, if we do that, we'd still end up with this logic duplicated for the editor and the renderer.
We can avoid that by creating a new component inside `packages/blocks/components/form-phone-numer`:

```JavaScript
/**
 * External dependencies
 */
import PhoneInput from 'react-phone-number-input';
import Input from 'react-phone-number-input/input';

const FormPhoneNumber = ( { flag, ...props } ) => {
	if ( flag ) {
		return <PhoneInput { ...props } />
	}

	return <Input { ...props } />
}

export default FormPhoneNumber

```

Make sure to export it in the `index.js` file.
And now, you can use it in the editor and in the renderer:

```JavaScript
	<FormPhoneNumber
		flag={ flag }
		...
	/>
```

package.json
```diff
@@ -25,7 +25,8 @@
   "dependencies": {
     "@emotion/cache": "^11.7.1",
     "@emotion/react": "^11.9.0",
-    "@emotion/styled": "^11.8.1"
+    "@emotion/styled": "^11.8.1",
+	"libphonenumber-js": "1.10.13"
```
jcheringer, line 29: Are we using this dependency?

jcheringer (changes requested):
Overall, it looks good.
I think we need some adjustments in the styles.
The input field should match the styles from the `Text Input Block` (colors, border, height) so we can design nice forms when placing them together.

image

The height handler isn't working well. And, when the flag dropdown is disabled, the input width isn't filling the available space.

image

package.json
DustinHartzler: No, we are not. Initially I was going to, removing it now.

packages/blocks/src/phone-number/index.js
DustinHartzler: Okay, perfect. This makes a lot of sense. Thanks!

packages/block-editor/src/phone-number/edit.js
DustinHartzler: Done.

packages/theme-compatibility/src/leven/base.scss
```diff
@@ -61,6 +61,7 @@ body {
 	}
 }
 
+.crowdsignal-forms-phone-number-block {
```
jcheringer, line 64: This line breaks the CSS compilation

packages/block-editor/src/phone-number/attributes.js
```diff
@@ -0,0 +1,40 @@
...
+		type: 'string',
+		default: '',
+	},
+	flag: {
+		type: 'boolean',
+		default: false,
+	},
+	label: {
+		type: 'string',
+		default: 'Phone Number',
```
jcheringer, line 16: This makes the block label have an actual value when placed instead of having a placeholder text.

packages/block-editor/src/phone-number/edit.js
```diff
@@ -0,0 +1,84 @@
...
+const EditPhoneNumber = ( props ) => {
+	const { attributes, setAttributes, className, isSelected } = props;
+
+	const flag = attributes.flag;
+
+	useClientId( props );
+
+	const handleChangeLabel = ( label ) => setAttributes( { label } );
+
+	// const handleChangeCountry = ( country ) => setAttributes( { country } );
```
jcheringer, line 28: Let's remove this comment

packages/block-editor/package.json
```diff
@@ -31,6 +31,7 @@
     "@wordpress/icons": "^9.3.0",
     "classnames": "^2.3.1",
     "lodash": "^4.17.21",
+	"react-phone-number-input": "^3.2.11",
```
jcheringer, line 34: Now we created a wrapper component inside the `blocks` package we don't need this dependency here.

packages/blocks/src/phone-number/index.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * External dependencies
+ */
+import { RawHTML } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import classnames from 'classnames';
+import { isEmpty } from 'lodash';
+import 'react-phone-number-input/style.css';
```
jcheringer, line 8: Nit: Can we put this under a `Styles dependencies` section?

packages/theme-compatibility/src/leven/base.scss
jcheringer: Also, we might need to increase the selector specificity.
It's still being overridden in the Renderer

jcheringer (changes requested):
Adding a few more clean-up comments.
I'm still seeing some inconsistencies regarding the block behavior in the Editor, especially regarding the ResizableBox.

image

Maybe we could just remove the ResizableBox?
Anyway, we need to make sure the input takes all the available space if the field width isn't defined.

packages/block-editor/src/phone-number/edit.js
DustinHartzler: Done!

packages/block-editor/src/phone-number/attributes.js
DustinHartzler: Changed to placeholder

packages/block-editor/package.json
DustinHartzler: When I removed it, the linter wouldn't let me commit. It said it was needed here.

packages/theme-compatibility/src/leven/base.scss
DustinHartzler: I've fixed up the CSS. Let me know if there is a better way to do this.

DustinHartzler:
> I'm still seeing some inconsistencies regarding the block behavior in the Editor, especially regarding the ResizableBox.
> Maybe we could just remove the ResizableBox?

I can remove the resizable box if that's the easiest solution.

jcheringer:
Hey @DustinHartzler, did you push your changes?
I do not see any commits after my last comments.

DustinHartzler:
@jcheringer - I think I got everything, but the default, non-country block, does not change with the 25% , 50%, etc picker. Any ideas?

packages/theme-compatibility/src/leven/base.scss
jcheringer: I saw you added a closing `}` to the end, but that's wrapping `.crowdsignal-forms-date-picker-block` and `.crowdsignal-forms-time-picker-block` and will break the Date/Time Picker.
To fix it, you need to replace the `.crowdsignal-forms-phone-number-block {` with `.crowdsignal-forms-phone-number-block,` and remove the closing `}` you added last.
In the end, it should look like this:

```CSS
.crowdsignal-forms-phone-number-block,
.crowdsignal-forms-date-picker-block .react-datepicker-wrapper,
.crowdsignal-forms-time-picker-block .react-datepicker-wrapper {
	input {
		background-color: var(--crowdsignal-forms-input-background-color);
		border: 1px solid var(--crowdsignal-forms-input-border-color);
	}
}
```

And regarding my comment to increase the selector specificity:
If you go to the Renderer, the block doesn't have a white background on the Leven theme.
To fix that, change the two `<PhoneInputWrapper>` on `packages/blocks/src/components/form-phone-number/index.js` to:

```JavaScript
<PhoneInputWrapper className="crowdsignal-forms-phone-input-wrapper">
```
And now, change the CSS to:

```CSS
.crowdsignal-forms-phone-number-block .crowdsignal-forms-phone-input-wrapper,
.crowdsignal-forms-date-picker-block .react-datepicker-wrapper,
.crowdsignal-forms-time-picker-block .react-datepicker-wrapper {
	input {
		background-color: var(--crowdsignal-forms-input-background-color);
		border: 1px solid var(--crowdsignal-forms-input-border-color);
	}
}
```

jcheringer:
> @jcheringer - I think I got everything, but the default, non-country block, does not change with the 25% , 50%, etc picker. Any ideas?

Add a `width: 100%` to the `input` on `packages/blocks/src/components/form-phone-number/index.js`:

```JavaScript
export const PhoneInputWrapper = styled.div`
	input {
		width: 100%;
	...	
```

packages/theme-compatibility/src/leven/base.scss
DustinHartzler: Thanks!

The first one was a silly mistake. The linter was complaining about a missing bracket and I didn't notice the extra one there.

And the second point makes total sense. I added it and now works as expected :)

jcheringer (approved):
LGTM!

PR #323, opened 2022-09-28
Update MC Question to allow 'other' answer

jcheringer:
## Summary

This PR updates the MC Question block to include the `Allow "other" answer` option.
The toggle was added to the block Sidebar.

image

## Testing Instructions
* Open/create a project
* Add a new MC Question block
* Open the block Sidebar and enable the "other" answer option
* Play with the block to ensure it works as expected

### DIFF
--- a/packages/block-editor/src/components/edit-button/index.js
+++ b/packages/block-editor/src/components/edit-button/index.js
@@ -31,13 +31,14 @@ const ButtonContent = styled.div`
 
 const EditButtonAnswer = ( {
 	value,
-	attributes,
+	attributes = {},
 	className,
 	multipleChoice,
 	onChange,
 	onReplace,
 	onSplit,
 	onDelete,
+	placeholder = __( 'Enter an answer', 'block-editor' ),
 	allowedFormats = [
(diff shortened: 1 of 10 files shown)


TheSteveK (changes requested):
If you change your mind and want to uncheck your "other" answer, you can't.

Also, you can check more boxes than the limit you set, and I was able to do so and submit without triggering an error.  I'm not sure if that was introduced in this or was always a bug in the block.

jcheringer:
> If you change your mind and want to uncheck your "other" answer, you can't.
> 
> Also, you can check more boxes than the limit you set, and I was able to do so and submit without triggering an error. I'm not sure if that was introduced in this or was always a bug in the block.

Managing the selection state for the "Other Answer" is a bit tricky.
I'm removing the checkbox/radio from the "Other Answer" and mimicking the behavior we have on legacy surveys.

Regarding the answer limit, that wasn't introduced by these changes, so I'd prefer handling it on a separate PR.

TheSteveK (approved):
Good enough for me :)

PR #324, opened 2022-09-29
Add time picker variation to date picker

TheSteveK:
The last of the blocks for CS-UI, the time picker block!  this is created as a variation of the date picker block.  It leverages the React-Datepicker's ability to show a time selector only, giving us a whole separate block.

To be fair I'm not 100% thrilled with the UI on the time picker (click on the clock icon once you click the time to get a set of dropdowns to adjust the time).  But, it's what the component has, and in the interest of speeding up deployment, I ran with it.

Screen Shot 2022-09-29 at 3 24 00 PM

Testing, you may need to run yarn-install again.  Open up the editor, add the time picker from the block library and off you go.  Add it to a project, respond and then check the results.  The patch for the date picker works for this too, so you don't need any backend patches.

### DIFF
--- a/packages/block-editor/src/date-time-picker/attributes.js
+++ b/packages/block-editor/src/date-time-picker/attributes.js
@@ -17,4 +17,20 @@ export default {
 	backgroundColor: {
 		type: 'string',
 	},
+	showTimeInput: {
+		type: 'boolean',
+		default: false,
+	},
+	showTimeSelectOnly: {
+		type: 'boolean',
+		default: false,
+	},
+	dateFormat: {
+		type: 'string',
+		default: 'MMMM d, yyyy',
+	},
+	placeholderText: {
+		type: 'string',
+		default: 'July 16, 1969',
+	},
 };
--- a/packages/block-editor/src/date-time-picker/edit.js
+++ b/packages/block-editor/src/date-time-picker/edit.js
@@ -45,7 +45,10 @@ const EditDateTimePicker = ( props ) => {
 					value={ attributes.label }
 				/>
 			</FormInputWrapper.Label>
-			<FormDatePicker placeholderText={ attributes.placeholderText } />
+			<FormDatePicker
+				placeholderText={ attributes.placeholderText }
+				showTimeSelectOnly={ attributes.showTimeSelectOnly }
+			/>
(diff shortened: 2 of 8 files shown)


jcheringer (changes requested):
In general, I think we need to improve a few things.

When you add the Time Picker, the label gets a text value instead of placeholder text:
image

When you select the block in the editor, it doesn't show the time picker, but it shows the arrow:
image


I think the UX for selecting a time isn't ideal. 
The library offers the [Time Only Option](https://reactdatepicker.com/#example-select-time-only), which I think looks better
image

I'm not sure if typing inside the input makes much sense since we don't offer any sort of masking. Maybe we should look for a way to disable typing on the input field.

Maybe using a variation for this block isn't the best idea.
We should be able to make it work using by controlling everything we need through attributes, but I'm not sure if that's the right approach.

Creating a new block would give us more flexibility to work without impacting the other block and would make the code a bit cleaner since each block would only keep the necessary attributes/props.

TheSteveK:
I hadn't noticed you can type in the input field.  That's also true for the date variation.  I think there's a way to replace the default input for a custom one.

As for doing its own block, my original plan was to roll my own time picker that was just based on input fields since we only need a text value for time.  But, after talking with @digitalwaveride there's a time crunch and we essentially need to have all of the CS UI stuff wrapped up by today.  For time I looked at just adding time into the date picker with block settings to enable it.  But, then it seemed like a variation would be a little better.

I could break out the variation into its own block, but I don't know if it can happen within the tight deadline.

TheSteveK:
Looking at it again, it does make sense to make a new block, but using the React-DatePicker component.  I'm starting that as a new PR.

jcheringer:
> Looking at it again, it does make sense to make a new block, but using the React-DatePicker component. I'm starting that as a new PR.

Yeah, that's what I have in mind.
Creating a new block but using the same library we used for the DataPicker.
Let me know if I can help you with anything.

TheSteveK:
I've pretty much got it done.  I'm currently trying to figure out if there's a way to prevent typing in the input box, though as a dropdown I don't think people will be as tempted.  I'm going to close this PR in favor of opening a new one for the dedicated block.

PR #325, opened 2022-09-30
Add dedicated time picker block

TheSteveK:
The last of the blocks for CS-UI, the time picker block! this is created as a variation of the date picker block. It leverages the React-Datepicker's ability to show a time selector only, giving us a whole separate block.

To be fair I'm not 100% thrilled with the UI on the time picker (click on the clock icon once you click the time to get a set of dropdowns to adjust the time). But, it's what the component has, and in the interest of speeding up deployment, I ran with it.

Testing, you may need to run yarn-install again. Open up the editor, add the time picker from the block library and off you go. Add it to a project, respond and then check the results. 

This will require D88920

Note that due to the way the source component handles the input box or custom input box, I don't think it's possible to keep people from typing in the field. But, since this uses the dropdown now, I don't see it as big of an issue.  And if you type in a valid time, it does work anyway.  If you don't, the component will reset the entry.


packages/blocks/src/time-picker/index.js
```diff
@@ -0,0 +1,68 @@
...
+		},
+	} );
+
+	const handleTimeChange = ( e ) => {
+		onChange( e.toTimeString() );
+	};
+
+	const parsedTime = useMemo( () => {
+		// the DatePicker component has to have a fully formatted date, adding an arbitrary date to value
+		return Date.parse( 'Thu Sep 29 2022 ' + fieldValue );
```
jcheringer, line 36: Since we just care about the time part of the string, there's no harm in using a hard-coded date. It's no a good practice, though.
If we change this logic to:

```JavaScript
	const parsedTime = useMemo( () => {
		if ( ! isEmpty( fieldValue ) ) {
			return  new Date( `${ new Date().toDateString() } ${ fieldValue }` );
		}
	}, [ fieldValue ] );
```

We get rid of the hard-coded string, and we also fix an issue where the initial value of the field is always set, instead of showing the placeholder.

packages/blocks/src/time-picker/index.js
```diff
@@ -0,0 +1,68 @@
...
+	ErrorMessage,
+	FormInputWrapper,
+	FormDateTimePicker,
+} from '../components';
+import { useField } from '@crowdsignal/form';
+
+const TimePicker = ( { attributes, className } ) => {
+	const { error, onChange, fieldValue } = useField( {
+		fieldName: `q_${ attributes.clientId }[text]`,
+		// initialValue: new Date().toDateString(),
```
jcheringer, line 22: Let's remove the commented line.

jcheringer (changes requested):
Looks good!
Adding small comments on things we could improve a bit.

jcheringer (approved):
Nice! :shipit:

PR #326, opened 2022-09-30
Enable Branding for Projects

jcheringer:
## Summary

This PR enables branding on Projects.
When creating projects, free users will see a sticky branding footer with a "Hide" button redirecting them to the Crowdsignal pricing page

Screen Shot 2022-09-30 at 16 12 34

On the Renderer, since it's intended to be accessed by the user's audience, the "Hide" button isn't shown.

Depends on: D88938-code

## Testing Instructions

* Apply backend changes (D88938-code)
* Using a free account, open/create a project
* You should see the branding on the footer of the Editor and the Renderer
* Change to a non-free account
* Now, you shouldn't see the branding

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -19,7 +19,7 @@
   },
   "main": "src/index.js",
   "dependencies": {
-    "@automattic/isolated-block-editor": "2.18.0",
+    "@automattic/isolated-block-editor": "2.18.1",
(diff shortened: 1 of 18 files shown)


jcheringer:
Converting this to Draft since we need https://github.com/Automattic/isolated-block-editor/pull/187 released for this to work properly.

PR #329, opened 2024-04-18
Allow defining a style nonce for the embed popup

jcheringer:
## Summary

This PR changes the `<crowdsignal-popup />` component to allow defining a `style-nonce` for environments with more restrictive CSP rules.

## Testing Instructions

The easiest way to test it is to build and reference the JavaScript file in a simple HTML file.
You can use the one here: `apps/embed/test.html.`

* Checkout this branch and run `yarn install & yarn build` in the root directory
* It should create the file `apps/embed/dist/main.js`
* Edit the [test HTML](https://github.com/Automattic/crowdsignal-ui/blob/main/apps/embed/test.html#L241-L242) file and change the reference from `https://app.crowdsignal.com/embed.js` to the recently built file (`apps/embed/dist/main.js`)
* Change the `<crowdsignal-popup />` and add a `style-nonce` and a valid `src`  property.
* Open the test HTML file and inspect the page source code.
* You should see the `nonce` attribute in all the `<style />` tags the `<crowdsignal-popup />` creates

### DIFF
--- a/apps/embed/src/popup.js
+++ b/apps/embed/src/popup.js
@@ -27,10 +27,17 @@ class CrowdsignalPopup extends window.HTMLElement {
 	 */
 	#settings;
 
+	/**
+	 * Allow defining a nonce for more environments with more restrictive CSP rules
+	 */
+	#styleNonce;
+
 	connectedCallback() {
 		const isWpEditor =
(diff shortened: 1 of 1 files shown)


jgcaruso (approved):
Nonce is added to style tags as described, but I only see 1 of the style tags rendered.

Is there something conditional that renders the other one?

image

PR #332, opened 2026-05-31
chore: pin marvinpinto/action-automatic-releases to v1.2.1 SHA

mahangu:
Pins `marvinpinto/action-automatic-releases@latest` to `d68defdd11f9dcc7f52f35c1b7c236ee7513bcc1`, the commit `@latest` currently resolves to (2022-01-29). Freezes the mutable tag without changing behavior.

Verify:

    gh api repos/marvinpinto/action-automatic-releases/commits/latest --jq .sha
    # => d68defdd11f9dcc7f52f35c1b7c236ee7513bcc1

Tracking: DEVPROD-1072.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 10
+    groups:
+      actions-minor-patch:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+      actions-major:
+        patterns:
+          - "*"
+        update-types:
+          - "major"
+    cooldown:
+      default-days: 7
+      semver-major-days: 14
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,4 +20,3 @@ updates:
           - "major"
     cooldown:
       default-days: 7
-      semver-major-days: 14
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -55,7 +55,7 @@ jobs:
                   cd ..
 
             - name: Create GH release
-              uses: marvinpinto/action-automatic-releases@latest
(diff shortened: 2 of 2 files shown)


donnchawp:
Reviewed this — the pin itself is fine and the SHA checks out (`919008cf` is genuinely the `v1.2.1` lightweight tag, and all the `with:` inputs still exist at that ref). Two notes:

**1. The description's premise is slightly off — this is a minor rollback, not a behavior-preserving pin.** `@latest` currently resolves to `d68defdd` (2022-01-29), which is 3 commits *ahead* of `v1.2.1` (2021-06-03). Those commits are entirely `dist/index.js` rebuilds, so the workflow was actually running a newer bundle than v1.2.1, and this pin rolls it back ~8 months. It's almost certainly safe — `make-release.yml` only runs on releases and v1.2.1 shipped fine historically — but the description says v1.2.1 is "the most recent," which isn't accurate. Suggest tweaking the wording to "deliberate pin to the last tagged release (accepting a minor rollback from the untagged `@latest` build)" so it reads correctly for anyone auditing DEVPROD-1072 later.

**2. Partial hardening.** `setup-node@v3.5.1` and `checkout@v2.5.0` in the same workflow are still on mutable version tags. If the goal is supply-chain posture, worth either SHA-pinning these too or explicitly noting they're deferred.

Minor FYI: the action is dormant (last release 2021-06-07, `runs.using: node12`), so Dependabot will never auto-bump it — the pin freezes it until someone migrates to a maintained alternative. Fine as a stopgap.

Dependabot config looks valid (`cooldown.default-days` is supported for the github-actions ecosystem, group split is sound).

mahangu:
Thanks for the careful look.

1. Good catch on the rollback. `@latest` resolves to `d68defdd`, which is 3 commits (all `dist/index.js` rebuilds) ahead of `v1.2.1`, so I've re-pinned to `d68defdd` to keep the current behavior instead of rolling back. The action's definition file (`action.yml`) is identical at both commits, so it still takes the same `with:` inputs you verified.

2. `setup-node` and `checkout` are GitHub-owned (`actions/*`), out of scope for this third-party pass. They get pinned in the first-party sweep (DEVPROD-1098) before enforcement.

Agreed it's a stopgap given the action's dormant, and thanks for confirming the Dependabot config.

donnchawp (approved):
Looks good!

